PR #16, opened 2015-01-08
Property Value Null Checks

codeclever:
I've updated the check on propertyValue to check for IsNullOrWhiteSpace() because GetPropertyValue
was returning an empty string rather than null therefore not using my AltPropertyName or the
DefaultValue if set.

Nice one,
Aaron ( - First pull request so go easy! Think I've done it right...)

### DIFF
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -185,7 +185,8 @@
                                                 : content.GetPropertyValue(umbracoPropertyName, recursive);
 
                         // Try fetching the alt value.
-                        if (propertyValue == null && !string.IsNullOrWhiteSpace(altUmbracoPropertyName))
+                        if ((propertyValue == null || propertyValue.ToString().IsNullOrWhiteSpace()) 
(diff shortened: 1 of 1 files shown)


leekelleher:
Thanks for the PR @codeclever! (I've been offline the past few days, with flu/fever - better now :-))

I'm 50/50 about merging in the PR... I'm thinking about the cases where you may expect the return value to be an actual empty string?

I'm not ruling it out, just want to cover the bases before merging it in.

// @mattbrailsford @JimBobSquarePants - second opinions? :-)

mattbrailsford:
I guess I'd go with whatever the default behavior in core is. If core uses the alt field on empty string, I'd probably to the same for constancy (I'm guessing they do what @codeclever is suggesting, but I'm not 100% sure).

leekelleher:
Thanks @mattbrailsford - I'll take a look at what `Umbraco.Field` does - makes sense!

leekelleher:
Yup, the legacy Umbraco `item` checks if the value is an empty string.
https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Web/umbraco.presentation/item.cs#L97

I'll merge this in - thanks again @codeclever, much appreciated!

PR #17, opened 2015-01-08
Move the context cache hack

JimBobSquarePants:
Moved so that non model-factory implementations can use the request
cache and included the type name in the cache key in case you are
returning a base type and inherited type within the same request.

I also shifted the type conversion code into it's own method as it would
have been a nightmare to read within the function.

Resharper 9 seems to have changed casing for excluded namespace folders so that's why the dotsettings file has been altered.

### DIFF
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -104,164 +104,197 @@
         /// <exception cref="InvalidOperationException">
         /// Thrown if the given type has invalid constructors.
         /// </exception>
-        internal static object As(
-            this IPublishedContent content,
-            Type type,
-            Action<ConvertingTypeEventArgs> convertingType = null,
-            Action<ConvertedTypeEventArgs> convertedType = null)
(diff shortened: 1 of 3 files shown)


leekelleher:
Thanks for the PR @JimBobSquarePants - as always, much appreciated.

Unfortunately, there is an underlying issue with caching the mapped models. See this thread for details, (the juicy stuff is on page 2):
http://our.umbraco.org/projects/developer-tools/ditto/ditto-feedback/57167-BaseController-BaseViewModel-with-Hijacked-controllers

Issue is with self-referential POCO mappings - e.g. a node having a property that will map to itself; like a menu/nav item.  Then because Ditto is trying to cache the model, it will try to use the same cache key.

Re: the other refactoring - looks good. I'm happy to merge those in - maybe a separate PR? (sorry if a pain)

JimBobSquarePants:
Interesting. So the caching is causing a problem?

I think that error message was the one I was getting when trying to use the PublishedContentFactory with
links.

I can mitigate a lot of the performance stuff with some of the tricks I have up my sleeve so I am happy to remove that bit and resubmit. I'll do that tonight.

JimBobSquarePants:
Closing as per request.

Issue #3, opened 2014-07-11
Reflection Caching?


leekelleher:
Consider caching the reflected objects.

We haven't looked into the performance of the reflection calls (e.g. `GetConstructors` or `GetProperties`), could it be improved?

Maybe a dictionary lookup instance?

    var cache = new Dictionary<Type, PropertyInfo[]>();

JimBobSquarePants:
I have a load of code already written that will cover caching so don't start anything.

leekelleher:
Cool! :sunglasses:

leekelleher:
@JimBobSquarePants Is this covered off in PR #19? Can I close this issue?

JimBobSquarePants:
Yeah I'd say so.

PR #19, opened 2015-01-21
Adding performance improvements

JimBobSquarePants:
Replaces the cache hack with private threadsafe dictionaries for storing
constructor and parameter info to reduce reflection overhead. This
should solve both issue #3 and also prevent circular referencing bugs
noted in the forums.

Additionally an extension method has been added for type initialisation.
This allow much faster creation of instance types than
Application.CreateInstance.

Refactorings are also present from other pull request.


leekelleher:
Cool, thanks @JimBobSquarePants! I'll merge it in.

PR #20, opened 2015-01-21
Use DependencyResolver to load TypeConverters

stevetemple:
This will allow the type converter implementations to use dependency injection, so they're more testable etc.

### DIFF
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -9,6 +9,7 @@
     using global::Umbraco.Core;
     using global::Umbraco.Core.Models;
     using global::Umbraco.Web;
+    using System.Web.Mvc;
 
     /// <summary>
     /// Encapsulates extension methods for <see cref="IPublishedContent"/>.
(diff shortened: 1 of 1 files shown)


leekelleher:
Thanks @stevetemple, very much appreciated!

I'm not too familiar with `DependencyResolver.Current.GetService` (I've read up a little, but never used it - yet).  Will it always create an instance of the type, or would we safer to add in a fall back condition for `Activator.CreateInstance`?

I'd love to have Ditto be testable - guess I'm looking for guidance in that area.

// @mattbrailsford @JimBobSquarePants Any feedback?

leekelleher:
@stevetemple Are you using this on a project? I'm interested in how the dependency injection works with the `TypeConverter`

stevetemple:
Yes we started to write our own and found yours so we've just started using it. If you don't have a IoC framework like Ninject etc installed it should fallback to `Activator.CreateInstance`

We've wrapped your code into a class because extension methods are tricky to Mock out in tests, so we've got a simple class with an interface wrapper so this can be mocked out.

JimBobSquarePants:
I've no experience `DependencyResolver` but I like the idea. I would need to see some example code of it used with `TypeConverter` in order to comment further though.

As for testing, Ditto certainly needs that. I can write tests for certain features if you set up a project but I don't know how to get everything working in order to mock `IPublishedContent` objects.

stevetemple:
I'll try and find a sample, basically in our type converter we needed to fetch media and other data so we injected something like IMediaService in the constructor of the TypeConverter so needed this to support it.

At the moment we haven't created any tests for Ditto itself, but it's something we could help with

leekelleher:
Thanks @stevetemple - I'll give it a test tomorrow morning, then I'll merge it in.

@JimBobSquarePants - thanks for feedback too!

I'll check the Umbraco Core unit-tests, as I'm sure that they have `IPublishedContent` mocks.

stevetemple:
The problem we have with IPublishedContent mocks is the extension methods like `GetPropertyValue` which as an extension method isn't mockable in Moq, but apparently something can be done in Fakes

leekelleher:
Bit delayed in my testing, all works fine with my existing project/solution - merged in.
Thanks again @stevetemple!

PR #21, opened 2015-01-21
Parse TypeConverter attributes on enumerables

JimBobSquarePants:
This addition allows us to detect custom TypeConverter attributes for enumerable types. This lets us add the TypeConverter attribute to the type argument class itself rather than each property which dramatically cuts down fluff. 

Adding the TypeConverter attribute to the property will override any attributes bound to classes.

E.g

```
[TypeConverter(typeof(FooConverter))]
public class Foo
{
    // Properties here
}

public class Bar
{
    // This property will automatically map
    public IEnumerable<Foo> Foos { get; set; }
}
```

### DIFF
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -271,9 +271,24 @@
                     // Process the value.
                     if (propertyValue != null)
                     {
-                        // Try any custom type converters first. Check both on class or property.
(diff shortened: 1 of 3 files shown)


leekelleher:
Cool, thanks @JimBobSquarePants.

With the new extension method classes, (`TypeInferenceExtensions` and `TypeInitializationExtensions`), I'm considering marking them as `internal` - is that cool with you?
Just that I don't want to expose too much API via Ditto.

JimBobSquarePants:
That's totally cool with me. I think the less exposed the better.

leekelleher:
Cool, all done in 53d2fbdd992758656981de63475a20dff6df8071 - cheers!

PR #22, opened 2015-01-23
Added support for Umbraco Dictionary Item values.

leekelleher:
Added support for Umbraco Dictionary Item values.

I'm not 100% happy about having a private static field for `UmbracoHelper`, but I didn't want to create a new instance for each property mapping - otherwise the Dictionary lookup may hit the database multiple times.

// @JimBobSquarePants - feel free to add suggestions

### DIFF
--- a/src/Our.Umbraco.Ditto/Attributes/UmbracoDictionaryValueAttribute.cs
+++ b/src/Our.Umbraco.Ditto/Attributes/UmbracoDictionaryValueAttribute.cs
@@ -1,0 +1,28 @@
+﻿namespace Our.Umbraco.Ditto
+{
+    using System;
+
+    /// <summary>
+    /// The Umbraco dictionary value attribute.
+    /// Used for providing Umbraco with additional information about a dictionary item to aid property value conversion.
+    /// </summary>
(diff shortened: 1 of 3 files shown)


JimBobSquarePants:
Maybe you could use an instance that is cached per request? I've used the following code in the past which seems to work well and remains performant. I have it sitting in a helper class that aids with conversion since you need an `UmbracoHelper` for quite a few of the TypeConverter implementations.

```
/// <summary>
/// Gets the <see cref="UmbracoHelper"/> for querying published content or media.
/// </summary>
public UmbracoHelper UmbracoHelper
{
    get
    {
        // Pull the item from the cache if possible to reduce the overhead caused by multiple reflection iterations 
        // for the given type taking place in a single request
        return (UmbracoHelper)ApplicationContext.Current.ApplicationCache.RequestCache.GetCacheItem(
                "UmbracoHelper",
                () => new UmbracoHelper(UmbracoContext.Current));
    }
}
```

leekelleher:
@JimBobSquarePants I'll merge in #24 before this one, then I can reuse the `ConverterHelper.UmbracoHelper` property.

PR #23, opened 2015-01-23
TypeInitializationExtensions - use DependencyResolver?

leekelleher:
@JimBobSquarePants - Following on from #20 - do you see any issue with us swapping out using `Activator` with `DependencyResolver`?

### DIFF
--- a/src/Our.Umbraco.Ditto/Extensions/TypeInitializationExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/TypeInitializationExtensions.cs
@@ -5,6 +5,7 @@
     using System.Linq;
     using System.Linq.Expressions;
     using System.Reflection;
+    using System.Web.Mvc;
 
     /// <summary>
     /// Extensions methods for <see cref="T:System.Type"/> for creating instances of types faster than 
@@ -21,7 +22,7 @@
         public static object GetInstance(this Type type)
         {
             // This is about as quick as it gets.
-            return Activator.CreateInstance(type);
+            return DependencyResolver.Current.GetService(type);
         }
 
         /// <summary>


JimBobSquarePants:
As log as it falls back to `Activator` as the default then I think it's fine. In 99% cases I very much doubt anyone would be using dependency injection.

On the note of `TypeConverter`s I wanted to as whether we should establish a collection of built in converters for the various Umbraco datatypes. I already have working code for most of the main ones.

leekelleher:
I agree, but now that the DI topic has been raised, it's how far do we go with it?

// @stevetemple - What are your thoughts on this? Is it an over-generalisation to be replacing every occurrence of `Activator.CreateInstance` with `DependencyResolver.Current.GetService`?

JimBobSquarePants:
My spelling in that last comment was appalling!

I'd like to also consider performance as a factor in feature design so any insight on the impact of the change from that perspective would be most useful.

leekelleher:
@JimBobSquarePants - It appears that the `GetService` of the default resolver (`DefaultDependencyResolver`) literally just calls `Activator.CreateInstance` (wrapped in a try/catch): 
[see here](https://github.com/ASP-NET-MVC/aspnetwebstack/blob/master/src/System.Web.Mvc/DependencyResolver.cs#L174).

re: Built-in TypeConverters (to ship with Ditto) ... sounds good, open a separate issue/PR.
Would this mean having built-in POCOs? (not as keen on that)

JimBobSquarePants:
@leekelleher That's good to know. I would definitely us `DependencyResolver` in that instance then but leave any other `Activator.CreateInstance` usages throughout the codebase intact since they don't involve anything that we'd want too many people messing with.

re: Built-in TypeConverters. No POCO's are used in any implementations I have done but I couldn't be sure about if we tried to do grid or related links.

stevetemple:
A good built in TypeConverter would be one for the new Grid stuff in 7.2

leekelleher:
OK, let's open a new PR and see where we go with built-in TypeConverters :muscle:

stevetemple:
One of the guys here is actually writing that grid `TypeConverter` already so we'll send that over once it's done

leekelleher:
Cool, thanks @stevetemple!

JimBobSquarePants:
Ace. I'll open a PR in an hour or so with a few.

PR #29, opened 2015-02-13
TypeConverters refactor [LK]

leekelleher:
@JimBobSquarePants I did some refactoring of the built-in TypeConverters. Mostly pulling in what I'm using on the Carlsberg projects.

I wanted to run them passed you via a PR, rather than commit directly.

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/HtmlStringConverter.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/HtmlStringConverter.cs
@@ -5,6 +5,7 @@
     using System.Globalization;
     using System.Web;
 
+    using global::Umbraco.Core.Dynamics;
     using global::Umbraco.Web.Templates;
 
     /// <summary>
@@ -27,7 +28,7 @@
         /// </returns>
         public override bool CanConvertFrom(ITypeDescriptorContext context, Type sourceType)
         {
-            if (sourceType == typeof(string) || sourceType == typeof(HtmlString))
+            if (sourceType == typeof(string) || sourceType == typeof(HtmlString) || sourceType == typeof(DynamicXml))
             {
                 return true;
             }
@@ -51,27 +52,34 @@
                 return null;
             }
 
-            var s = value.ToString();
-            if (value is HtmlString)
-            {
-                if (!string.IsNullOrWhiteSpace(s))
-                {
-                    s = TemplateUtilities.ParseInternalLinks(s);
-                }
-
-                return new HtmlString(s);
-            }
-
-            // This lets us convert textbox multiple data type properties to HtmlString's 
             if (value is string)
             {
-                if (!string.IsNullOrWhiteSpace(s))
+                var text = value.ToString();
+
+                if (!string.IsNullOrWhiteSpace(text))
                 {
                     var umbracoHelper = ConverterHelper.UmbracoHelper;
-                    s = umbracoHelper.ReplaceLineBreaksForHtml(s);
+                    text = umbracoHelper.ReplaceLineBreaksForHtml(text);
                 }
 
-                return new HtmlString(s);
+                return new HtmlString(text);
+            }
+
+            if (value is HtmlString)
+            {
+                var html = value.ToString();
+
+                if (!string.IsNullOrWhiteSpace(html))
+                {
+                    html = TemplateUtilities.ParseInternalLinks(html);
+                }
+
+                return new HtmlString(html);
+            }
+
+            if (value is DynamicXml)
+            {
+                return ((DynamicXml)value).ToHtml();
             }
 
             return base.ConvertFrom(context, culture, value);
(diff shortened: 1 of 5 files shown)


src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/HtmlStringConverter.cs
```diff
@@ -51,27 +52,34 @@ public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo c
...
+                    html = TemplateUtilities.ParseInternalLinks(html);
                 }
 
-                return new HtmlString(s);
+                return new HtmlString(html);
+            }
+
+            if (value is DynamicXml)
+            {
+                return ((DynamicXml)value).ToHtml();
```
JimBobSquarePants: Would this need parsing for links?

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/MediaPickerConverter.cs
```diff
@@ -48,24 +49,47 @@ public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo c
...
+            if (id <= 0)
+            {
+                return null;
+            }
+
+            var umbracoHelper = ConverterHelper.UmbracoHelper;
+            var media = umbracoHelper.TypedMedia(id);
+
+            // Ensure we are actually returning a media file.
+            if (media.HasProperty(Constants.Conventions.Media.File))
```
JimBobSquarePants: Didn't know about this. I like it. :+1:

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/MultiNodeTreePickerConverter.cs
```diff
@@ -56,7 +56,13 @@ public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo c
             if (!string.IsNullOrWhiteSpace(s))
             {
                 var multiNodeTreePicker = Enumerable.Empty<T>();
-                var nodeIds = s.Split(new[] { "," }, StringSplitOptions.RemoveEmptyEntries).Select(int.Parse).ToArray();
+
+                int n;
+                var nodeIds = s
+                    .ToDelimitedList()
```
JimBobSquarePants: Again. Cool!

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/HtmlStringConverter.cs
leekelleher: Not sure, I'll review my original code - see what happens with the property data.

JimBobSquarePants:
Looks good to me. I say go for it.

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/MediaPickerConverter.cs
leekelleher: I painstakingly added all those to Umbraco core :satisfied:

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/MultiNodeTreePickerConverter.cs
leekelleher: Since finding `ToDelimitedList`, I seem to be using it *everywhere* :smile_cat:

leekelleher:
Ace! Thanks for reviewing it.

JimBobSquarePants:
No worries :smile:

mattbrailsford:
Suggestion for multi node pickers. Check for a Max count of 1 in prevalue and return as IPublishedContent rather than enumerable.

leekelleher:
@mattbrailsford Good point!  Though accessing the prevalues would hit the database right? (I'd like to avoid that.)

Hmmm, I wondering if we should take the approach that a TypeConverter should only return one `Type`.  This would mean reviewing the `MediaPickerConverter` class, as that can return a single media item or (if it's a folder) child items.

mattbrailsford:
The problem with type converters is that umbraco only ever asks what you
can convert from, not to, so you have to just make your best match. This
raises the issue though that if we only return one type and had two type
converters, only the first would match as they both match on the "can
convert from" request.

I guess it would be nicer for umbraco to also check the "can convert to"
methods of type converters to make sure the right one is used.

Matt

leekelleher:
Again, good points @mattbrailsford! I agree, defo room for improvement.

On an aside, you've just reminded me that in Ditto, when we look for a `TypeConverterAttribute` (at property level), we only call the `ConvertFrom` method... but we should also check `CanConvertFrom`.
(I'll go fix that)

mattbrailsford:
Good catch!

leekelleher:
I've just checked the Umbraco core source, and `TryConvertTo` handles both `CanConvertTo` and `CanConvertFrom` checks: [see here in ObjectExtensions.cs](https://github.com/umbraco/Umbraco-CMS/blob/Release-4.10.0/src/Umbraco.Core/ObjectExtensions.cs#L84). It's been there since v4.10.0.

I'm still trying to get my head around what the difference between `ConvertTo` and `ConvertFrom` actually are... whenever I think I've got it, I'm like ooooh :shit:!

mattbrailsford:
Convert from is based on the stored values type. Convert to is based on the
outputs value type.

From is usually going to be a string or stored into id or something where
as To is likely to be something complex.

PR #31, opened 2015-02-13
Call CanConvertFrom when a custom TypeConverterAttribute

leekelleher:
I've opened this as a PR, as I have one question... see inline comment below.

### DIFF
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -378,8 +378,11 @@
                         else if (propertyInfo.PropertyType == typeof(HtmlString))
                         {
                             // Handle Html strings so we don't have to set the attribute.
-                            HtmlStringConverter converter = new HtmlStringConverter();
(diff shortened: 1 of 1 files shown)


src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -344,7 +344,7 @@ private static object GetTypedProperty(IPublishedContent content, Type type)
                                     if (toConvert != null)
                                     {
                                         var converter = DependencyResolver.Current.GetService(toConvert) as TypeConverter;
-                                        if (converter != null)
+                                        if (converter != null && converter.CanConvertFrom(propertyType))
```
leekelleher: Question: What should we be passing to `CanConvertFrom`? `propertyType` or `propertyValue.GetType()`?

e.g. The intended type, or the type of the current value?

leekelleher:
I think I've just answered my own question... it should be `propertyValue.GetType()`.

mattbrailsford:
It should be the type of the current value.

leekelleher:
Thanks @mattbrailsford. I realised that too when I was typing the question. Doh! :-)

You've got me thinking more about how `CanConvertTo` be used.

Issue #26, opened 2015-02-04
v6 Support


JimBobSquarePants:
I'm wondering if we could do a Ditto.Compat project for v6? A lot of agencies I know are still using that major version.

I have a load of code written already that will allow v6 support for type conversion and run a custom version of the project that used conditional compilation to spit out the correct code using `#if` 

Obviously there are issues with using the built in generic methods like `Ancestors<T>` since we have no `PublishedContentFactory` but as far as I'm aware that affects v7 also if the factory is not used. Am I correct in that assumption?

I have a base class that contains strong typed equivalents for those methods e.g.  `GetAncestors<T>` just now. I wonder whether we can simply give it the same name... There might be ambiguous reference issues though.

We would have to use `PublishedContentExtended` rather than `PublishedContentModel` for any inheritance but thats not an issue.

What are your thoughts?

leekelleher:
We only compiled against v7.1.4 since that's when the `IPublishedContentModelFactory` was introduced.  To be honest, I'm less convinced that it's a good fit for Ditto - I've considered dropping support for it, (or maybe move to a separate assembly?)

I have no problem with v6 support - it is totally possible. Just that I didn't want the responsibility of testing it (lazy me :anguished:). My current projects are v7.1+, so was easier for me to test against.

re: `Ancestors<T>` - this isn't used in Ditto, is it?

JimBobSquarePants:
I'm happy to test it. I use it all the time. :smile: 

We can keep it in a separate project and include a lot of the files as references from the main project. I'll open a branch of develop to do so. 

Dropping the `IPublishedContentModelFactory` support? I like it. If you use `Ancestors<T>` etc with the factory enabled I'm sure it automatically maps to your type. I'd have to double check some of my code to be sure.

leekelleher:
`IPublishedContentModelFactory` could be moved to a separate assembly? (a la `Ditto.PublishedContentModelFactory` - *gawd that's a mouthful!*)

Ditto doesn't use `Ancestors<T>` verbatim - the model-factory is called whenever the Umbraco core requests an `IPublishedContent` node (internally it checks if a factory is set, then calls `CreateModel()`).

leekelleher:
With v6 support... I'm wondering how far back can we support? (I think `IPublishedContent` was introduced in v4.10.) or is that too far?

leekelleher:
So Ditto could have 2 assemblies:

[...]

PR #32, opened 2015-02-13
Moved the ModelFactory code to its own project

leekelleher:
Referenced in #26

Doing this helps us remove the constraint from being tied to Umbraco v7.1.4 - we can add support for v6.x (maybe even v4.10.x?)


JimBobSquarePants:
Beat me to the punch! This was gonna be part of v6 support. Will help muchly!

leekelleher:
Thought you'd said "fire ahead"? (in #26) ;-)

JimBobSquarePants:
Hehehe... Ace. :+1:

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/MultiNodeTreePickerConverter.cs
```diff
@@ -58,7 +60,9 @@ public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo c
                 var multiNodeTreePicker = Enumerable.Empty<T>();
 
                 int n;
-                var nodeIds = s
+                var nodeIds = XmlHelper.CouldItBeXml(s)
+                    ? uQuery.GetXmlIds(s)
```
leekelleher: Thinking about my uQuery concern, (since I know it will be removed in Umbraco v8.0), I tried to find an alternative API within the Umbraco, but no luck!

I reckon that we should include the `GetXmlIds()` as an internal helper method (within Ditto). Would `ConverterHelper` be an appropriate place to have it?

src/Our.Umbraco.Ditto/packages.config
```diff
@@ -1,25 +1,23 @@
...
+  <package id="Microsoft.AspNet.WebApi.WebHost" version="4.0.30506.0" targetFramework="net45" />
   <package id="Microsoft.AspNet.WebPages" version="2.0.20710.0" targetFramework="net45" />
   <package id="Microsoft.Web.Infrastructure" version="1.0.0.0" targetFramework="net45" />
   <package id="MiniProfiler" version="2.1.0" targetFramework="net45" />
   <package id="MySql.Data" version="6.6.5" targetFramework="net45" />
-  <package id="Newtonsoft.Json" version="6.0.2" targetFramework="net45" />
+  <package id="Newtonsoft.Json" version="4.5.11" targetFramework="net45" />
   <package id="SharpZipLib" version="0.86.0" targetFramework="net45" />
-  <package id="UmbracoCms.Core" version="7.1.4" targetFramework="net45" />
+  <package id="UmbracoCms.Core" version="6.2.5" targetFramework="net45" />
```
leekelleher: I'm in 2 modes of thought about this:

1. We could support much further back than this - say v4.10.x
2. We only support the latest, as if you are using Ditto, you'd generally be doing so from the start of a project, not trying to retro-fit it?

@mattbrailsford @JimBobSquarePants @Jeavon - Any thoughts?

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/MultiNodeTreePickerConverter.cs
leekelleher: For ref, the code for `uQuery.GetXmlIds()` is here:
https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Web/umbraco.presentation/umbraco/uQuery/uQuery.cs#L143

src/Our.Umbraco.Ditto/packages.config
JimBobSquarePants: Converter helper would be a fit I think, so yeah, let's put it in there. 

As far as support..I think as far back in 6 as possible, I just chose the latest for speeds sake when building.

Going any further back could cause issues with API's. We need as much raw speed as possible out of this so hacking round things could prove difficult.

leekelleher: I think we'll stick with Umbraco v6.2.5 for our next release.

Going forwards, I've posted a question on the forum:
https://our.umbraco.org/projects/developer-tools/ditto/ditto-feedback/61876-Ditto-Earliest-version-of-Umbraco-to-support
Let's see _if_ anyone responds :hear_no_evil:

Jeavon: I would stick with latest supported versions of Umbraco (currently v6 and v7). I think it would be pretty odd to start a new project using Umbraco v4 and need Ditto....

leekelleher: I agree!

No one has responded on the forum, but that's due to this :smile_cat: 
meow

Jeavon: Is that cat actually alive?

leekelleher: Noooo... doing that to a live cat would just be plain cruel!

Jeavon: Agreed :dizzy:

Issue #27, opened 2015-02-05
Conflicting assemblies when building against Umbraco 7.2.1


iainjames88:
When building against a fresh install of Umbraco 7.2.1 (with no other packages installed) Visual Studio warns:

***"1>C:\Program Files (x86)\MSBuild\12.0\bin\Microsoft.Common.CurrentVersion.targets(1697,5): warning MSB3277: Found conflicts between different versions of the same dependent assembly that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed."***

The conflicting assemblies might be System.Web.Mvc and Umbraco?

leekelleher:
Hi @iainjames88, I've had chance to look at this, but unfortunately I couldn't reproduce the error/warning message.

I set-up a clean Umbraco 7.2.1 site (via NuGet). Check that it installed okay, then went back to Visual Studio (2012) and installed Ditto 0.5.0 (via NuGet). I recompiled the solution and didn't have any errors/warnings.  The Umbraco web-app (front-end/back-office) worked okay too.

Are you still experiencing this?

leekelleher:
@iainjames88 - Well, good news - I've managed to reproduce the issue. Bad news - I've only just hit it and haven't fixed it yet.  I'm actively looking into a fix - I'll keep you posted via this issue.

leekelleher:
OK, I think I've fixed it - released as [Ditto 0.5.1](https://www.nuget.org/packages/Our.Umbraco.Ditto/0.5.1) ... if you want to test it @iainjames88?

leekelleher:
Closing this issue, as believed to be fixed in [v0.5.1](https://github.com/leekelleher/umbraco-ditto/releases/tag/0.5.1)

PR #33, opened 2015-02-18
Hotfix 0.5.1

leekelleher:
Reverted the `System.Web.Mvc` dependency from v4.0.0.1 back to v4.0.0.0.
As this is causing issues with environments that have not yet been patched with v4.0.0.1 yet.

This patch should resolve issue #27.


leekelleher:
Oops, targeted `develop` instead of `master` - doh!

PR #35, opened 2015-03-03
Adds a context to the conversions.

JimBobSquarePants:
This commit add passing the current IPublishedContent as a context to
TypeConverters.

With this addition it's now possible to use `context.Instance` within custom `TypeConverters`

    public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
    {
        IPublishedContent content
        if (context != null)
        {
            content = (IPublishedContent)context.Instance;
        }
		
		// Do the conversion.
    }

This is great if you need to substitute values when converting. E.g Substituting `SeoTitle` with `Name`

### DIFF
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -4,6 +4,7 @@
     using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.ComponentModel;
+    using System.Globalization;
     using System.Linq;
     using System.Reflection;
     using System.Web;
@@ -355,9 +356,7 @@
                                             // Create context to pass to converter implementations.
                                             // This contains the IPublishedContent and the currently converting property name.
                                             var context = new PublishedContentContext(content, actualPropertyName);
-
-                                            // TODO: Should we pass the current culture when converting?
-                                            object converted = converter.ConvertFrom(context, null, propertyValue);
+                                            object converted = converter.ConvertFrom(context, CultureInfo.CurrentCulture, propertyValue);
 
                                             // Handle Typeconverters returning single objects when we want an IEnumerable.
                                             // Use case: Someone selects a folder of images rather than a single image with the media picker.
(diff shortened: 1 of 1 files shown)


leekelleher:
Good job piecing this together!  Now I get that question on StackOverflow :+1: 

I'll get some time later to review and merge in.

mattbrailsford:
Nice :)

JimBobSquarePants:
Cheers. :smile:  Yeah, I knew what I wanted to do just didn't have enough WinForms experience playing about with TypeConverters to be absolutely sure I was correct. I think this really fleshes out conversion now to make Ditto really robust.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -346,12 +352,18 @@ private static object GetTypedProperty(IPublishedContent content, Type type)
                                         var converter = DependencyResolver.Current.GetService(toConvert) as TypeConverter;
                                         if (converter != null)
                                         {
+                                            // Create context to pass to converter implementations.
+                                            // This contains the IPublishedContent and the currently converting property name.
+                                            var context = new PublishedContentContext(content, actualPropertyName);
+
+                                            // TODO: Should we pass the current culture when converting?
+                                            object converted = converter.ConvertFrom(context, null, propertyValue);
+
```
leekelleher: Yeah, I think it would be useful to pass in the current culture, (it's better than having a `null` right?)

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -379,7 +391,8 @@ private static object GetTypedProperty(IPublishedContent content, Type type)
                         {
                             // Handle Html strings so we don't have to set the attribute.
                             HtmlStringConverter converter = new HtmlStringConverter();
-                            propertyInfo.SetValue(instance, converter.ConvertFrom(propertyValue), null);
+                            var context = new PublishedContentContext(content, actualPropertyName);
+                            propertyInfo.SetValue(instance, converter.ConvertFrom(context, null, propertyValue), null);
```
leekelleher: Note: Same for here, pass in the current culture.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
JimBobSquarePants: You wanna add it or shall I?

leekelleher: Go for it :+1: (It's your feature :-))

leekelleher:
@JimBobSquarePants Once you've added in the current culture code, feel free to merge this into `develop`, it's a great feature for the next release.

JimBobSquarePants:
Will do. :smile:

JimBobSquarePants:
Done. Whoop! :metal:

PR #41, opened 2015-03-05
Adds a `CultureInfo` parameter to the `.As<T>` extension methods.

leekelleher:
Enabling the current culture to be passed into the conversion.


leekelleher:
@JimBobSquarePants - Can you review this please?

Not sure whether it should be part of 0.6.0 or wait until next release? Might as well do 0.6.0 if you're happy?

PR #42, opened 2015-03-05
Adding support for SymbolSource for both NuGet and MyGet

Jeavon:
This will cause the build scripts to generate both a .nupkg and a symbols.nupkg.

The nupkg will contain exactly the same contents as it does currently, the symbols.nupkg will additionally contain the pdb file and all cs files, this package will be automatically pushed to symbolsource.org (no config needed) and Visual Studio will be able to debug if set to use symbol servers!

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,6 +18,7 @@ deploy:
      # MyGet Deployment for builds & releases
   - provider: NuGet
     server: https://www.myget.org/F/umbraco-ditto/
+    symbol_server: https://nuget.symbolsource.org/MyGet/umbraco-ditto
     api_key:
       secure: Q1/4K8VSwr7BjwmKDTef8y5lOc7S+jK9ELuWy67y6OVRpjxmnF9M3Gfs1kT+ir8x
     artifact: /.*\.nupkg/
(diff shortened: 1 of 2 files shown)


leekelleher:
MIND BLOWN

Jeavon:
Nice! I didn't think it would be this easy when I started looking into it for @warrenbuckley but it is!

JimBobSquarePants:
HOW DO YOU KNOW THIS STUFF????

Jeavon:
I've no idea!

brain

leekelleher:
@Jeavon I'll merge this in... quick check, do I need to register on symbolsource.org? (For any reason?)

Jeavon:
No you don't to sign up, they hooked up with MS/NuGet. You can if you want to see stats etc... You have to add your NuGet API key.

The MyGet packages on Symbolsource are also hooked up and you will already have a login/password which is the same as your MyGet log in. https://www.symbolsource.org/MyGet/

As far as I've been able to tell you can't join up a NuGet and MyGet accounts on SymbolSource, you just end up with the same one twice, once for /MyGet and one for /Public

leekelleher:
OK cool - so, so far you've only added this for the MyGet deploy right?

Jeavon:
NuGet is fully automated due to the hook up with MS, so this will work for both now

leekelleher:
Groovy, all merged in!

Thanks again @Jeavon! #h5yr

Jeavon:
Just grabbed the artefacts and the both look perfect, the magic of NuGet!
image

PR #44, opened 2015-03-08
Adding archetype support to Ditto

micklaw:
Adds support for archetype via the existing UmbracoPropertyAttributeClass which now contains an isArchetype field.

This hooks in to the existing TypeConverters and flow of application. Archetypes POCOs should follow the same naming convention standards as the main ditto POCOs, though will inherit from a base class of ArchetypeModel instead of the IPublishedContent.

I have ran this against almost all of the existing data types and conversion seems OK, though as archetype values are strings one or two data types may have slipped through.

### DIFF
--- a/src/Our.Umbraco.Ditto/Attributes/UmbracoPropertyAttribute.cs
+++ b/src/Our.Umbraco.Ditto/Attributes/UmbracoPropertyAttribute.cs
@@ -24,12 +24,13 @@
         /// <param name="defaultValue">
         /// The default value.
         /// </param>
-        public UmbracoPropertyAttribute(string propertyName, string altPropertyName = "", bool recursive = false, object defaultValue = null)
+        public UmbracoPropertyAttribute(string propertyName, string altPropertyName = "", bool recursive = false, object defaultValue = null, bool isArchetype = false)
         {
             this.PropertyName = propertyName;
             this.AltPropertyName = altPropertyName;
             this.Recursive = recursive;
             this.DefaultValue = defaultValue;
+            this.IsArchetype = false;
         }
 
         /// <summary>
@@ -51,5 +52,10 @@
         /// Gets or sets the default value.
         /// </summary>
         public object DefaultValue { get; set; }
+
+        /// <summary>
+        /// Defines if property is an Archetype
+        /// </summary>
+        public bool IsArchetype { get; set; }
     }
 }
--- a/src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/MultiNodeTreePickerConverter.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/MultiNodeTreePickerConverter.cs
@@ -100,7 +100,9 @@
                 return multiNodeTreePicker;
             }
 
-            return base.ConvertFrom(context, culture, value);
+            // [ML] - return default list as its always nice to avoid null reference exceptions on lists =)
+
+            return Enumerable.Empty<T>();
         }
 
         /// <summary>
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -1,4 +1,9 @@
-﻿namespace Our.Umbraco.Ditto
+﻿using System.Collections;
+using Newtonsoft.Json;
+using Our.Umbraco.Ditto.Models.Archetype;
+using Our.Umbraco.Ditto.Models.Archetype.Abstract;
+
+namespace Our.Umbraco.Ditto
 {
     using System;
     using System.Collections.Concurrent;
@@ -19,6 +24,9 @@
     /// </summary>
     public static class PublishedContentExtensions
     {
+        private static readonly ConcurrentDictionary<string, Type> ArchetypeCache
+           = new ConcurrentDictionary<string, Type>();
+
         /// <summary>
         /// The cache for storing constructor parameter information.
         /// </summary>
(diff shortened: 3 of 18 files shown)


src/Our.Umbraco.Ditto/Our.Umbraco.Ditto.csproj
```diff
@@ -32,6 +32,16 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="interfaces">
```
JimBobSquarePants: Minimum version has to be 6.2.5

JimBobSquarePants:
That's a big commit! I'll try and set aside some time tonight to have a look at it though I see @leekelleher get's up earlier than me on a Sunday so he'll probbaly beat me to it. :smile:

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/MultiNodeTreePickerConverter.cs
```diff
@@ -100,7 +100,9 @@ public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo c
                 return multiNodeTreePicker;
             }
 
-            return base.ConvertFrom(context, culture, value);
+            // [ML] - return default list as its always nice to avoid null reference exceptions on lists =)
```
JimBobSquarePants: I dunno about this. You're might be right but Microsoft guidelines indicate always using base converters as a fallback. https://msdn.microsoft.com/en-us/library/ayybcxe5.aspx

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -1,4 +1,9 @@
-﻿namespace Our.Umbraco.Ditto
+﻿using System.Collections;
+using Newtonsoft.Json;
+using Our.Umbraco.Ditto.Models.Archetype;
+using Our.Umbraco.Ditto.Models.Archetype.Abstract;
```
JimBobSquarePants: I got stung by this on my first commit. We're keeping a flat namespace structure. `Our.Umbraco.Ditto`

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -216,6 +224,294 @@ public static class PublishedContentExtensions
...
+        /// </param>
+        /// <param name="propertyInfo"></param>
+        /// The <see cref="PropertyInfo"/> of the property to set on the object
+        /// <param name="content">
+        /// The <see cref="IPublishedContent"/> of which to use as a base for populating content
+        /// </param>
+        /// <param name="isArchetype">
+        /// A <see cref="bool"/> specifying if is of type archetype
+        /// </param>
+        private static void SetValue(Type type, object instance, object propertyValue, PropertyInfo propertyInfo, IPublishedContent content = null, bool isArchetype = false)
```
JimBobSquarePants: I like that this is split out. Nesting was starting to get confusing.

src/Our.Umbraco.Ditto/Extensions/TypeInferenceExtensions.cs
```diff
@@ -90,5 +92,33 @@ public static IEnumerable<Type> GetBaseTypes(this Type type)
...
+        /// <param name="className">Name of the class sought.</param>
+        /// <param name="baseType"></param>
+        /// <returns>Types that have the class name specified. They may not be in the same namespace.</returns>
+        public static IEnumerable<Type> GetTypeByName(string className, Type baseType)
+        {
+            /* [ML] - Look away kids, forgive me. We could improve performance with an optional app setting to filter assemblies 
+             * but this method executes in low milliseconds, hopefully with caching of these results we'll be ok
+             */
+
+            var assemblies = AppDomain.CurrentDomain.GetAssemblies().Where(i => !i.FullName.ToLower().StartsWith("system") && !i.FullName.ToLower().StartsWith("microsoft") && !i.FullName.ToLower().StartsWith("windows")).ToList();
```
JimBobSquarePants: What's this bit loading up?

micklaw: As we need to grab an instance of the the archetype model POCO(s) to populate, we can only get this via the archetype alias (string), this method checks the dlls classes in the application for types which inherit from the base Archetype Model and where the class name matches the alias. Thinking about it we may also need to check if the alias has been set via an attribute, but it could be easily extended for that. 

I have filtered out System.*, Microsoft.* and Windows.* dlls as I am pretty sure we would want to exclude them.

The results of this are cached and also in testing with a test install the execution time for sub 30ms, so seemed like a hit worth taking on first load of each type. 

Cheers

leekelleher:
Thanks for the pull request!

I'll review in more detail tomorrow, (as I'm out & about today).

To be upfront, I doubt we'll merge this in as is, as it ties Ditto too closely to Archetype. But I definitely think there should be a Ditto-like way for mapping Archetype data.  Let's use this as a discussion point to move it forward. :+1:

src/Our.Umbraco.Ditto/Extensions/TypeInferenceExtensions.cs
JimBobSquarePants: Ah... Umbraco already has a method that can provide inherited types. I would imagine (I haven't checked the source) that it's highly optimised.

PluginManager.Current.ResolveTypes<YOUR_BASE_TYPE>()

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/MultiNodeTreePickerConverter.cs
micklaw: Cool, I think the base converters throw exceptions as values passed in are always strings from archetype, so some are unable to cast to the new type a flake out. Sure there will be some work around we can use.

micklaw:
No worries, just for clarity, this amend doesn't have any dependency to archetype at all, so even if it's not in the solution this should still function alright.

src/Our.Umbraco.Ditto/Extensions/TypeInferenceExtensions.cs
micklaw: Good to know, cheers for that.

leekelleher:
@micklaw Thanks again for your PR! I'm reviewing it now :-)

Given the size of the commit, I'll probably ask a lot of questions... don't feel disheartened if I nitpick ;-)

src/Our.Umbraco.Ditto/Attributes/UmbracoPropertyAttribute.cs
```diff
@@ -51,5 +52,10 @@ public UmbracoPropertyAttribute(string propertyName, string altPropertyName = ""
         /// Gets or sets the default value.
         /// </summary>
         public object DefaultValue { get; set; }
+
+        /// <summary>
+        /// Defines if property is an Archetype
+        /// </summary>
+        public bool IsArchetype { get; set; }
```
leekelleher: I think this binds Ditto too much with Archetype (not in terms of assembly reference, but conceptually).

Should we consider other 3rd party property-editors with complex data? `IsNuPicker`, `IsSirTrevor`, `IsVorto`
It makes Ditto non-generic.

Let's look for an alternative solution - preferably using TypeConverters.

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/MultiNodeTreePickerConverter.cs
leekelleher: I have the same concern as @JimBobSquarePants - having the base converter as fallback.

@micklaw - Yup, I'm sure we can workaround issues in the body of the `ConvertFrom`.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
leekelleher: :+1: Flat namespace.  It's totally subjective ;-)

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
leekelleher: :+1: I'm good with this too.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -216,6 +224,294 @@ public static class PublishedContentExtensions
...
+                if (isArchetype)
+                {
+                    /* [ML]: TODO - There must be a better way to do this to avoid an archetype dependency with its model binding to CMS json. */
+
+                    var json = propertyValue is string ? propertyValue as string : JsonConvert.SerializeObject(propertyValue, new JsonSerializerSettings
+                    {
+                        ReferenceLoopHandling = ReferenceLoopHandling.Ignore
+                    });
+
+                    var jsonModel = JsonConvert.DeserializeObject<ArchetypeCMSModel>(json);
```
leekelleher: :warning: This raises alarm bells for me... as it appears to bypass Archetype's own PropertyValueConverter.
(I may be wrong, I'm still reviewing)

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -216,6 +224,294 @@ public static class PublishedContentExtensions
...
+                    {
+                        if (fieldset.Properties != null && fieldset.Properties.Any())
+                        {
+                            Type instanceType = null;
+
+                            ArchetypeCache.TryGetValue(fieldset.Alias, out instanceType);
+
+                            if (instanceType == null)
+                            {
+                                instanceType = TypeInferenceExtensions.GetTypeByName(fieldset.Alias, typeof(ArchetypeModel)).FirstOrDefault();
```
leekelleher: Trying to understand what this does. Is it looking for a class/type that matches the Archetype's fieldset's alias?

I'm concerned that we're not going through Archetype's own `GetValue<T>` method:
https://github.com/imulus/Archetype/blob/master/app/Umbraco/Umbraco.Archetype/Models/ArchetypePropertyModel.cs#L32
... as this holds context of the underlying property-editor's type... and associated `PropertyValueConverter`s.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -216,6 +224,294 @@ public static class PublishedContentExtensions
...
+
+                                    AddToPropertyCache(instanceType);
+                                }
+                            }
+
+                            if (instanceType != null)
+                            {
+                                // ML - Create an instance for each archetype object
+
+                                var instance = Activator.CreateInstance(instanceType) as ArchetypeModel;
```
leekelleher: OK, so we're asking that the POCO has to inherit from `Our.Umbraco.Ditto.Models.Archetype.Abstract.ArchetypeModel`?

:warning: _I'd prefer not to do enforce this constraint._

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -296,162 +587,39 @@ public static class PublishedContentExtensions
...
                         defaultValue = umbracoPropertyAttr.DefaultValue;
+                        isArchetype = umbracoPropertyAttr.IsArchetype;
                     }
 
                     // Try fetching the value.
                     var contentProperty = contentType.GetProperty(umbracoPropertyName);
-                    object propertyValue = contentProperty != null
-                                            ? contentProperty.GetValue(content, null)
-                                            : content.GetPropertyValue(umbracoPropertyName, recursive);
+                    object propertyValue = contentProperty != null ? contentProperty.GetValue(content, null) : content.GetPropertyValue(umbracoPropertyName, recursive);
```
leekelleher: The original whitespace/formatting was intended, (and totally subjective)

src/Our.Umbraco.Ditto/Extensions/TypeInferenceExtensions.cs
```diff
@@ -8,7 +10,7 @@
     /// Extensions methods for <see cref="T:System.Type"/> for inferring type properties.
     /// Most of this code was adapted from the Entity Framework
     /// </summary>
-    internal static class TypeInferenceExtensions
+    public static class TypeInferenceExtensions
```
leekelleher: Why mark as `public`? (we'd prefer not to - makes it difficult to manage the inner-workings in future)

If it is for the unit-tests project, then we can use the `InternalsVisibleTo` attribute:

    [assembly: InternalsVisibleTo("Our.Umbraco.Ditto.Tests")]

src/Our.Umbraco.Ditto/Models/Archetype/Abstract/ArchetypeModel.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Our.Umbraco.Ditto.Models.Archetype.Abstract
+{
+    public class ArchetypeModel
```
leekelleher: :warning: As mentioned before, I have concerns about Ditto (core) being too aware of Archetype - especially with being responsible for deserializing it's JSON, and bypassing it's PropertyValueConverter.

micklaw:
Ha, no worries, I'm thick skinned ;)

At the GLUUG a few of the guys thought it was a good idea. I had already
done something similar and thought it may be useful. I mentioned to
@JimBobSquarePants <https://github.com/JimBobSquarePants> I'd fire
something over. I always thought the solution to this problem was always
going to be subjective, though the idea and principle seemed like a good
one to me.



On 9 March 2015 at 10:30, Lee Kelleher <notifications@github.com> wrote:

> @micklaw <https://github.com/micklaw> Thanks again for your PR! I'm
> reviewing it now :-)
>
> Given the size of the commit, I'll probably ask a lot of questions...
> don't feel disheartened if I nitpick ;-)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/leekelleher/umbraco-ditto/pull/44#issuecomment-77830168>
> .
>



-- 

Cheers

Michael Law

leekelleher:
OK, hopefully I wasn't too much of an arsehole in my review?

In concept, I am 100% sure that we should find a way to have Ditto-like way for mapping Archetype data!

With the implementation, (as I keep going on about), I don't want the Ditto (core) to be aware of Archetype itself.  I think there is a way by using TypeConverters - which I'd like us to explore.

:heart: Thanks again for the PR - much kudos to you! (For taking the time to get involved, it is appreciated!) I'll buy you a :beer:!

src/Our.Umbraco.Ditto/Attributes/UmbracoPropertyAttribute.cs
micklaw: @leekelleher Type converters makes sense to me, could potentially extend the ArchetypeModel (https://github.com/imulus/Archetype/blob/master/app/Umbraco/Umbraco.Archetype/Models/ArchetypeModel.cs) with a FieldSetAs<T> method of something? Would these TypeConverters be completed in a separate assembly to avoid assembly dependency?

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
micklaw: @leekelleher I deserialized and reserialized to my own class to avoid a dependency on the Archetype dll. Umbraco seems to find the ArchetypeModel within the Archetype.dll as the property value so didn't want to pass this about.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
micklaw: @leekelleher Yeah, that's exactly what its doing.

Though the code doesn't explicitly use the PropertyValueConverters here, it does fall back to using this in the SetValue method contained within the class, the same way ditto does as its final fallback binding attempt.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
micklaw: @leekelleher Maybe an interface could be defined, in the same we you guys use IPublishedContent, I have found in our use cases that having base properties e.g Alias has come in handy.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
micklaw: @leekelleher Ha, clearly not using my portrait monitor that day =)

src/Our.Umbraco.Ditto/Extensions/TypeInferenceExtensions.cs
micklaw: @leekelleher nUnit trade off, always have an inner battle between testing only public methods versus wanting to ensure things like this work. InternalsVisibleTo makes me weep a little inside.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
leekelleher: Touché! :smiley_cat:

src/Our.Umbraco.Ditto/Extensions/TypeInferenceExtensions.cs
leekelleher: General thinking is that we don't expose too much (internals) from the Ditto core.
I'd hate for us to refactor code in future that other developers _might_ come to use/rely on.

src/Our.Umbraco.Ditto/Attributes/UmbracoPropertyAttribute.cs
leekelleher: :+1: That's where my thinking is at the moment - separate assembly to specifically to handle Archetype data.

micklaw:
@leekelleher @JimBobSquarePants Fellas, how would you feel if i refactored this code to be in the as we discussed in a seperate assembly and use a ArchetypeConverter<T> to bind the data?

Had a look over again and don't think it would much longer.

leekelleher:
@micklaw - I think there is a definite need for Archetype POCO mapping.

Discussing with @mattbrailsford about it being part of the Ditto project, (see my comment on #46), we came to the conclusion that it is beyond the scope of Ditto - in terms of both concept and ongoing support.

Saying that, I don't want to discourage you (or anyone else) from developing a solution... so yes, please feel free to do it as a separate project.

@mattbrailsford @JimBobSquarePants  - did you want to add comment on this?

micklaw:
@leekelleher Based on the previous pull request, I separated some of the logic to its own method for setting the value of the instance properties. I have taken this a step further and exposed the method in the class below as its own interfaced class with a single SetValue method.

https://github.com/micklaw/umbraco-ditto/blob/archetype/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs

As i see it in my tests, this would be the only change to the ditto source to make this class and interface public. The rest can be hooked in a separate assembly via type converters. This would mean the archetype integration could be completely separate from the ditto source.

micklaw:
Here's a sample project separated out, didn't take long to hook up. Core has no dependency on archtypes or anything else.

https://github.com/micklaw/umbraco-ditto/tree/archetype-refactor

jbreuer:
Hey did you read this topic about Archetype? https://github.com/zpqrtbnk/Zbu.ModelsBuilder/issues/13 Not really related to this, but more people want to generate models for Archetype ;-).

micklaw:
Hey, I defo think this has its place too. I'm a bit swamped for time at the
moment with other work but when i get some free time i think i'm going to
look in to creating an external assembly for Ditto which could potentially
map Archetype and Grid Type with POCOs.

On 7 May 2015 at 12:56, Jeroen Breuer <notifications@github.com> wrote:

> Hey did you read this topic about Archetype? zpqrtbnk/Zbu.ModelsBuilder#13
> <https://github.com/zpqrtbnk/Zbu.ModelsBuilder/issues/13> Not really
> related to this, but more people want to generate models for Archetype ;-).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/leekelleher/umbraco-ditto/pull/44#issuecomment-99831813>
> .
>



-- 

Cheers

Michael Law

leekelleher:
Closing this PR.

Thanks again @micklaw - stellar effort here! Noticed your new project/repo at [Ditto.Resolvers](https://github.com/micklaw/Ditto.Resolvers), I think that's the best move going forwards. (I may even contribute to it! :sunglasses:)

micklaw:
No worries man, I think your spot on with the separate repository approach. I really like the new changes in the develop branch, defo makes the framework much more flexible =) Good work dudes.

PR #45, opened 2015-03-08
Add MemberPicker TypeConverter.

leekelleher:


JimBobSquarePants:
I was waiting for someone to build this. Not sure why I missed it.

leekelleher:
I noticed it when working on the dictionary item converter.  I think it might need refactoring, as there's a better member object to use (extends IPublishedContent)

JimBobSquarePants:
Is there? I've not done enough membership stuff to know. I'll need to in the next couple of weeks though so this is good timing.

leekelleher:
Hmmm... I think I may have confused myself, I was thinking of the `Umbraco.Web.Models.ProfileModel` - but that seems to only apply to the currently logged in member.  Whereas when querying members you get an `IPublishedContent`.

See this doc for reference: https://our.umbraco.org/documentation/Reference/Querying/MemberShipHelper/#ProfileModelandIPublishedContent

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/MemberPickerConverter.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                return null;
+            }
+
+            if (value is int)
+            {
+                return this.ConvertFromInt((int)value);
+            }
+
+            var content = value as IPublishedContent;
```
leekelleher: @JimBobSquarePants - Quick question, I copied `ContentPickerConverter.cs` for this class, I'm wondering about this line - why do a cast first? Would it be better to do `if (value is IPublishedContent)` check first? _Just curious_

leekelleher:
So given `UmbracoHelper.TypedMember` returns an `IPublishedContent`, I think I'll merge this in :-)

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/MemberPickerConverter.cs
JimBobSquarePants: In a v7 world IPublishedContent would come first, in v6 it will be an int. 

Since v6 is legacy we should probably switch those round to improve performance.

leekelleher: Just to clarify, switch it so that we check for `IPublishedContent` first or `int` first?

PR #47, opened 2015-03-10
Add Enum converter.

JimBobSquarePants:
Adds support for converting to Enums. WIll convert from multiple types and is suitably robust. Can also be inherited from easily for Nupickers etc.


leekelleher:
Cool! Good addition! Thanks @JimBobSquarePants

PR #49, opened 2015-03-15
Allow TypeConverters to be called when property value is null

Hendy:
allows a TypeConverter to be executed for a null, such as when a value hasn't been set in Umbraco, or for when a property doesn't actually store a value (eg. nuPickers: Relation Labels)

### DIFF
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -389,7 +389,7 @@
                                             else
                                             {
                                                 // Return single expected items from converters returning an IEnumerable.
-                                                if (converted.GetType().IsEnumerableType())
+                                                if (converted != null && converted.GetType().IsEnumerableType())
                                                 {
                                                     // Generate a method using 'FirstOrDefault' to convert the type back to T.
                                                     MethodInfo firstMethod = typeof(Enumerable)
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -330,43 +330,49 @@
                     }
 
                     // Process the value.
-                    if (propertyValue != null)
+                    var propertyType = propertyInfo.PropertyType;
+                    var typeInfo = propertyType.GetTypeInfo();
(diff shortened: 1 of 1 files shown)


leekelleher:
@JimBobSquarePants Could you review the code amends please?

I'm good with the suggestion (of allowing TypeConverters on null values), but it looks like there are other minor code amends in this PR, (around lines 355, 370 and 440)

JimBobSquarePants:
I'll have a look. :+1:

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -330,43 +330,49 @@ public static class PublishedContentExtensions
...
-                                        // This contains the IPublishedContent and the currently converting property descriptor.
-                                        var descriptor = TypeDescriptor.GetProperties(instance)[propertyInfo.Name];
-                                        var context = new PublishedContentContext(content, descriptor);
+                                        propertyValueType = propertyValue.GetType();
+                                    }
 
-                                        if (converter.CanConvertFrom(context, propertyValue.GetType()))
-                                        {
-                                            object converted = converter.ConvertFrom(context, culture, propertyValue);
+                                    if (converter.CanConvertFrom(context, propertyValueType))
```
JimBobSquarePants: All the TypeConverters would have to be amended to allow `null` in the `CanConvertFrom` method.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -421,34 +427,40 @@ public static class PublishedContentExtensions
...
-                                propertyInfo.SetValue(instance, converter.ConvertFrom(context, culture, propertyValue), null);
-                            }
+                        Type propertyValueType = null;
+                        if (propertyValue != null)
+                        {
+                            propertyValueType = propertyValue.GetType();
                         }
-                        else if (propertyInfo.PropertyType.IsInstanceOfType(propertyValue))
+
+                        if (converter.CanConvertFrom(propertyValueType))
```
JimBobSquarePants: HtmlTypeConverter would have to be amended to accept null

JimBobSquarePants:
I'm gonna merge this in and make the TypeConverter changes myself as they are simple enough. It looks good to me.

leekelleher:
Cool, thanks! Go ahead with the merge :+1:

PR #51, opened 2015-03-16
Refactoring PublishedContentExtensions for readability and performance.

JimBobSquarePants:
The methods in this class were getting too difficult to follow and any changes difficult to detect. Resharper was having a hard time also.

I've split the `GetPropertyValue` method into two parts: a getter and a setter. I've also cached the `MethodInfo` instances for `Cast` and `FirstOrDefault` as static fields since they will never change.That should give us a performance booster.

@leekelleher @mattbrailsford Can you both cast your eyes over this when you can as I want to make sure I haven't introduced any bugs?

### DIFF
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -439,8 +439,8 @@
                                         if (!converted.GetType().IsEnumerableType())
                                         {
                                             // Generate a method using 'Cast' to convert the type back to IEnumerable<T>.
-                                            CastMethod.MakeGenericMethod(parameterType);
-                                            object enumerablePropertyValue = CastMethod.Invoke(null, new object[] { converted.YieldSingleItem() });
+                                            var cast = CastMethod.MakeGenericMethod(parameterType);
+                                            object enumerablePropertyValue = cast.Invoke(null, new object[] { converted.YieldSingleItem() });
                                             propertyInfo.SetValue(instance, enumerablePropertyValue, null);
                                         }
                                         else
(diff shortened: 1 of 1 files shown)


src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -292,182 +312,201 @@ public static class PublishedContentExtensions
...
-                                                    object singleValue = firstMethod.Invoke(null, new[] { converted });
-                                                    propertyInfo.SetValue(instance, singleValue, null);
-                                                }
-                                                else
-                                                {
-                                                    propertyInfo.SetValue(instance, converted, null);
-                                                }
-                                            }
+                                            // Generate a method using 'FirstOrDefault' to convert the type back to T.
+                                            FirstOrDefault.MakeGenericMethod(propertyType);
```
mattbrailsford: Is this right? Does MakeGenericMethod not return a new MethodInfo instance? and it's this instance we should call Invoke on?

JimBobSquarePants: Crap, Yup. You're right there, that compile but doesn't run you can perform late bound operations on a generic method like that. I'll fix.

mattbrailsford: Seems to be the case with uses of CastMethod too.

leekelleher:
Cool, thanks @JimBobSquarePants!

I wont get chance until later on to test this out, (damn we need `IPublishedContent` unit-tests! :wink:).
In the meantime, @mattbrailsford if you're happy with it - I'm happy for it to be merged in (to `develop`).

JimBobSquarePants:
No problemo.. :sunglasses: 

Yeah, we need those tests. I've never done any unit testing against `IPublishedContent` though so I don't know where to start.

PR #53, opened 2015-03-17
null check fix and fallback value for setting the current Culture

Hendy:
noticed error during a data import process (using Ditto outside of a normal page request)

### DIFF
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -243,9 +243,11 @@
                 {
                     culture = UmbracoContext.Current.PublishedContentRequest.Culture;
                 }
-
-                // fallback
-                culture = CultureInfo.CurrentCulture;
+                else
+                {
+                    // fallback
+                    culture = CultureInfo.CurrentCulture;
+                }
             }
 
             // Get the default constructor, parameters and create an instance of the type.
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -236,10 +236,10 @@
             Type type,
             CultureInfo culture = null)
         {
-            // Check if the culture has been set, otherwise use from Umbraco.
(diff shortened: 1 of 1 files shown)


src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -239,7 +239,13 @@ public static class PublishedContentExtensions
...
             if (culture == null)
             {
-                culture = UmbracoContext.Current.PublishedContentRequest.Culture;
+                if (UmbracoContext.Current.PublishedContentRequest != null)
+                {
+                    culture = UmbracoContext.Current.PublishedContentRequest.Culture;
+                }
+
+                // fallback
+                culture = CultureInfo.CurrentCulture;
```
JimBobSquarePants: Are we not missing an `else` here?

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -239,7 +239,15 @@ public static class PublishedContentExtensions
             // Check if the culture has been set, otherwise use from Umbraco.
             if (culture == null)
             {
-                culture = UmbracoContext.Current.PublishedContentRequest.Culture;
+                if (UmbracoContext.Current.PublishedContentRequest != null)
```
leekelleher: We'd probably need to check if `UmbracoContext.Current != null` too - otherwise may throw null reference.

Hendy: @leekelleher what should it fall back to is that's null ?

leekelleher: Sorry, I didn't mean as an extra if/else, but to add to the condition:

    if (UmbracoContext.Current != null && UmbracoContext.Current.PublishedContentRequest != null)

As if `UmbracoContext.Current` was null, then checking `UmbracoContext.Current.PublishedContentRequest` would throw the null reference.

leekelleher:
@JimBobSquarePants Assigning to you, as this may impact your refactoring PR... feel free to merge in once you're happy :-)

PR #54, opened 2015-03-19
Implement new build naming & numbering

Jeavon:
Six digit leading zero build numbers

Build numbers suffixed with a hyphen, e.g. "0.7.0-alpha-000022"

CI builds always have a suffix, by default "alpha" but could also be "beta"

If using "beta" CI build and then want to move to RTM, can now use special "rtm" suffix, set UMBRACO_PACKAGE_PRERELEASE_SUFFIX = rtm, when you release via lightweight tag, one build will be the release and will be successful, the other will intentionally fail to avoid a invalid CI build.

This allows this full release pattern if you want it:

	version: 0.7.0.{build}
	UMBRACO_PACKAGE_PRERELEASE_SUFFIX=

		CI > 0.7.0-alpha-0012
		CI > 0.7.0-alpha-0013

	UMBRACO_PACKAGE_PRERELEASE_SUFFIX=alpha
	APPVEYOR_REPO_TAG=true

		Release > 0.7.0-alpha

	UMBRACO_PACKAGE_PRERELEASE_SUFFIX=beta

		CI > 0.7.0-beta-0015

	APPVEYOR_REPO_TAG=true

		Release > 0.7.0-beta

	UMBRACO_PACKAGE_PRERELEASE_SUFFIX=rtm
	APPVEYOR_REPO_TAG=true
	(this is where a CI build would normally be created as well as the release, it would be 0.7.0-alpha-0017, but instead it will be aborted)

		Release > 0.7.0

Currently Ditto has only CI builds and releases so nothing needs to be changed.

When switching to this new numbering I would recommend that any existing MyGet packages for v0.7.0 are deleted from the feed so that the order starts to work in Visual Studio.

### DIFF
--- a/build/tools/AppVeyorUmbraco/AppVeyorUmbraco.Targets
+++ b/build/tools/AppVeyorUmbraco/AppVeyorUmbraco.Targets
@@ -17,8 +17,7 @@
             var pos = BuildVersion.LastIndexOf('.');
             var len = BuildVersion.Length - pos - 1;
             
-            var buildNumber = ("00000" + BuildVersion.Substring(pos + 1, len));
-            var buildNumberWithZeros = buildNumber.Substring(buildNumber.Length - 6);
+            var buildNumberWithZeros = BuildVersion.Substring(pos + 1, len).PadLeft(6, '0');
 
             var baseVersion = BuildVersion.Substring(0, pos);
 
--- a/build/tools/AppVeyorUmbraco/AppVeyorUmbraco.Targets
+++ b/build/tools/AppVeyorUmbraco/AppVeyorUmbraco.Targets
@@ -4,21 +4,44 @@
     TaskFactory="CodeTaskFactory"
     AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.v4.0.dll" >
 	<ParameterGroup>
-      <BuildVersion ParameterType="System.String" Required="true" />
+    <BuildVersion ParameterType="System.String" Required="true" />
 	  <BuildSuffix ParameterType="System.String" Required="false" />
 	  <ProductVersion ParameterType="System.String" Output="true" />
+  	<Release ParameterType="System.String" Required="true" />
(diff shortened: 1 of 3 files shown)


build/tools/AppVeyorUmbraco/AppVeyorUmbraco.Targets
```diff
@@ -4,21 +4,44 @@
...
-            var baseVersion = BuildVersion.Substring(0, BuildVersion.LastIndexOf('.'));
-            if (string.IsNullOrEmpty(BuildSuffix)){
-			        ProductVersion = baseVersion;
-            } else {
-              ProductVersion = baseVersion + "-" + BuildSuffix;
+            var pos = BuildVersion.LastIndexOf('.');
+            var len = BuildVersion.Length - pos - 1;
+            
+            var buildNumber = ("00000" + BuildVersion.Substring(pos + 1, len));
+            var buildNumberWithZeros = buildNumber.Substring(buildNumber.Length - 6);
```
leekelleher: Thinking you could replace these lines with a `PadLeft(6, '0')`?

e.g.
```csharp
var buildNumberWithZeros = BuildVersion.Substring(pos + 1, len).PadLeft(6, '0');
```

Jeavon: Yes, nice method I didn't know existed!

leekelleher:
Thanks @Jeavon - great work again!

It's up to you if you want to use that `PadLeft` instead... I'm happy to merge in whenever :+1:

Jeavon:
Done, once merged you will need to delete the existing v0.7.0 CI builds from MyGet

leekelleher:
Cool, all done - MyGet tidied up too :sunglasses:

Jeavon:
Fab, will make updating much easier! Thanks!

Jeavon:
Need to get Umbraco using this scheme now!

PR #55, opened 2015-03-19
Remove type parameters.

### LINKED ISSUE #52: Remove the need for <T> in converters.
Following a chat with @Hendy I've come to realise that we don't actually need the `<T>` type parameters on the TypeConverters we have them on any more since we pass the `PropertyDescriptor` within our context.

I guess that was the original intent @Hendy yeah?

We can use `context.PropertyDescriptor.ComponentType` to get the parent class and property type `context.PropertyDescriptor.PropertyType ` and the 
`As(this IPublishedContent content,Type type...)` extension to process that type. 

This would make using the built in converters a lot less verbose.

Now obviously this is a breaking change so that's why I've raised this for discussion. 

I quite like the idea of the user simple not caring about stuff like that since we have already dealt with it but I appreciate also that this will cause existing code to require updating and make our codebase a little bit more complicated.

What do you think?

JimBobSquarePants:
Removed all instances of T from the type converters. 

Fixes #52 

This was a biggie @Hendy! 

Additionally I've added high speed Enumerable invocation methods and ensured that the culture is passed throughout where it wasn't before.

I've done as much as I can to keep performance as high as possible here, I'd really like you guys to test it   out on your machines though as using runtime types rather than compile time types will always be slower.

The invocation methods are an order of magnitude faster than what was being used before though.


leekelleher:
@JimBobSquarePants It all looks good, but I'd like a chat with you, me & @mattbrailsford soon about it.

I'm all for performance improvements, etc - just that with some of the refactorings, I'm feeling a little overwhelmed (probably just due to gaps in my knowledge) ... but leads me to the question, if we make the code too clever, would we put off potential contributions?

Let's chat next week sometime? :surfer:

JimBobSquarePants:
I'll admit, some of that code was the most difficult stuff I have ever had to wrap my head round so I totally get your point. 

I think there are parts of Ditto, and any project really that most folk will be able to contribute to easily enough but but others that are more hardcore. ImageProcessor for example has some devilishly difficult stuff but is mostly trivial.

Those difficult bits though, I find anyway, are generally the least likely to change.

Performance, in my opinion is super important for Ditto. We need to api to fly as well as be simple to use or else there'll never really be real world use case for it. 

Brace yourself for when I upload the proxy stuff anyway... It's scary as hell to look at but so, powerful and useful.

leekelleher:
@mattbrailsford - I've had chance to take a closer look at the changes and they aren't as scary as I initially thought.  I'm happy to merge them in, but I'd like you to cast an eye over it before I do.

mattbrailsford:
Ok, I think you are right, it's not as scary as I thought. I think it's the stuff in the Common folder that scares me, but the use of those methods seems to be restricted to the core .As method, so shouldn't affect people contributing TypeConverters which was my concern (don't want type converters to become so complex people are scared to contribue them as this to me is a prime area for help from others).

1 thing that has struck me though is my dislike of the ConverterHelper class. This sounds like it's going to become a prime dumping ground for a mish-mash of various methods. My initial thought is can we make a base TypeConverter to extend from that exposes the UmbracoHelper (Maybe UmbracoTypeConverter to suggest there is some degree of Umbraco context-ness). And I'd move the GetXmlIds into a string extension method or something (ParseXmlIds?).

Just my 2 cents.

JimBobSquarePants:
I can go with that. There's a fair whack of repetitive code there.

Have you guys seen my comments regarding TypeConverters and model binding on Gitter. I'm concerned about our converters causing issues with form posting.

mattbrailsford:
Might be worth raising an issue RE model binders. Easier to keep a track of. Would need to see an example though ideally.

leekelleher:
@mattbrailsford re: `ConverterHelper` - now that you've brought it up, I'm liking the idea of a base `UmbracoTypeConverter` abstract class. :new_moon_with_face:

mattbrailsford:
I dun thunk something clever?

leekelleher:
_Thunk?_ Do you know something about [this](http://stackoverflow.com/questions/27190555/what-is-jscript-thunk-assembly-assembly)? :confused:

JimBobSquarePants:
@leekelleher @mattbrailsford Yo! 

As you can see I've been busy this evening. Everything in this branch now incorporates all the additions and fixes from develop plus the type param removal and other fixes.

Could you have a look over and give me your blessing to merge into develop?

leekelleher:
@JimBobSquarePants Excellent! :metal: 

Wondering about the `DittoBaseConverter` name, whether it should have "Umbraco" in there too? (As @mattbrailsford mentioned it provides Umbraco context) ... maybe `DittoUmbracoConverter` ... _urgh, sounds like a mouthful doesn't it?_

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/DittoBaseConverter.cs
```diff
@@ -0,0 +1,175 @@
...
+    /// <summary>
+    /// The Ditto base converter containing reusable properties and converting methods for <see cref="IPublishedContent"/> instances.
+    /// All other Ditto converters should inherit from this class.
+    /// </summary>
+    public abstract class DittoBaseConverter : TypeConverter
+    {
+        /// <summary>
+        /// Gets the <see cref="UmbracoHelper"/> for querying published content or media.
+        /// </summary>
+        public static UmbracoHelper UmbracoHelper
```
leekelleher: General question, is it better to use a `static` property or for it to be part of the instance?

leekelleher: Quick follow up, I was looking at how Umbraco core do it for their base/abstract classes...

* [UmbracoController.cs#L30](https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Web/Mvc/UmbracoController.cs#L30)
* [UmbracoControl.cs#L40](https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Web/UI/Controls/UmbracoControl.cs#L40)
* _(there are a few others)_

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/DittoBaseConverter.cs
```diff
@@ -0,0 +1,175 @@
...
+        {
+            get
+            {
+                // Pull the item from the cache if possible to reduce the db access overhead caused by 
+                // multiple reflection iterations for the given type taking place in a single request.
+                return (UmbracoHelper)ApplicationContext.Current.ApplicationCache.RequestCache.GetCacheItem(
```
leekelleher: I've still got this niggley feeling that we shouldn't really be using the `RequestCache` to store `UmbracoHelper`.  I'm not suggesting to change this, just wanted to flag it up for discussion.

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/DittoBaseConverter.cs
JimBobSquarePants: I guess we should probably do it per instance really. That way it is lazily invoked.

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/DittoBaseConverter.cs
JimBobSquarePants: Originally I did it as multiple type converters could be running on the same request. I honestly don't know what happens when an UmbracoHelper is created so I wanted to avoid any db driven or heavy going code.

JimBobSquarePants:
Also... Not precious in the slightest about the base converter name but I would imagine most people would know that Ditto is synonymous with Umbraco.

mattbrailsford:
RE base converter name, IMO no need for "base" to be in their as the class is abstract, so can't be anything but a base class. 

Also, I think the reason you are struggling to choose a name is because it's doing too many varied things. Essentially the dumping ground has been moved.

2 ways I would suggest to fix it.

1) Move the general helper methods that can into extension methods (which I think they were previously, so dunno why they moved) and just have the base class expose things like UmbracoContext that don't make sense as extension methods

2) Seperate the base class into 2 base classes, DittoConverter, which holds all the general helper methods, and UmbracoConverter (which extends DittoConverter) which exposes all things Umbraco.

If you have to explain what a class by saying "it does this AND this AND this" it's doing too much. You could say be able to say "it dos this".

DittoConverter - Gives access to general helpers methods / properties
UmbracoConverter - Gives access to umbraco helpers methods / properties

Just my 2 cents.

I also agree that if we have methods on the base classes, they shouldn't be static as it prevent people from overriding them if they want to. They should be protected virtual.

JimBobSquarePants:
Ok... Removing "Base" makes sense. It should be DittoUmbracoConverter if we are maintaining convention with other classes.

There's only one method that isn't Umbraco specific in the base converter class.

```
IsNullOrEmptyString(object value)
```

So I'm not sure about splitting classes up. I don't recall any extension methods for anything that is in that class either. 

:+1: for protected virtual though. Not sure what I was thinking.

JimBobSquarePants:
Do you think we should maybe pass the UmbracoHelper though as part of the context?

JimBobSquarePants:
I've done all the changes we discussed other than removing the UmbracoHelper.

The HtmlConverter uses the UmbracoHelper method `ReplaceLineBreaksForHtml(string text)` which internally uses the global configuration which we can't access. I don't know any workarounds for that.

I'm keen to get this all pushed to develop asap so Ill need help if we are to make those final changes.

leekelleher:
From look at the source for [`ReplaceLineBreaksForHtml`](https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Web/UmbracoHelper.cs#L1007) - it initially appears to be a simple check for rendering out "XHTML", e.g. a leading whitespace before the closing slash.

Since `Umbraco.Core.Configuration.GlobalSettings` is marked as internal, we can't get to it... but closer inspect of the [`EditXhtmlMode`](https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Core/Configuration/GlobalSettings.cs#L535) property... it's not only marked as obsolete, it also *always* returns `true`! :astonished:

Which means that we could just use `text.Replace("\n", "<br/>\n");` directly in the `HtmlConverter`?

Failing that, there is a static method in [`umbraco.library.ReplaceLineBreaks`](https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Web/umbraco.presentation/library.cs#L993), (that gets it's own `UmbracoHelper`) and then calls `ReplaceLineBreaksForHtml`.

JimBobSquarePants:
Wizzard! Aces, I'll move that to an extension method then.

leekelleher:
Cool, then all good to merge in :+1:

JimBobSquarePants:
I've removed the UmbracoHelper. I quick sense check please then I'll do the merge. :smile:

leekelleher:
I was going to question why move the extension methods to `Internal` folder, but think it makes sense... ;-)

All good... MERGE IT IN BABY! :shipit:

PR #57, opened 2015-04-01
Ditto.ModelFactory NuGet package

### LINKED ISSUE #43: Create NuGet package for Ditto.ModelFactory

leekelleher:
Creates NuGet package for Ditto.ModelFactory.

Closes #43


leekelleher:
The `package.proj` build script is starting to feel quite long.
I'm considering splitting it up, so that the ModelFactory build is kept separate.

// @Jeavon What are your thoughts on this? Better to have an MSBuild script for each project/package?

Jeavon:
@leekelleher yes, I would go with a different MsBuild Script for each NuGet package being published.

leekelleher:
Cool! :100: I'll fix it up with separate MSBuild scripts :+1:

leekelleher:
I've separated out the MSBuild scripts for each project/package.  I wasn't sure about the naming, so have called them `Ditto.proj` and `Ditto.ModelFactory.proj` for now.

PR #59, opened 2015-04-05
Update type converters to treat emtpy string the same as null

### LINKED ISSUE #58: MultiNodeTreePickerConverter<T> throws when passed an empty string
This pertains to Umbraco 7.2.4 and Ditto 0.6.1

Hello! I've been playing with integrating Ditto into a test project and ran into an issue when using the MultiNodeTreePickerConverter: it would throw when passed an empty string, which I believe happens when the user hadn't selected any content.

```
public class Event : BasePage
 {
    public Event(IPublishedContent content) : base(content)
    {
    }

    public DateTime StartTime { get; set; }
    public DateTime EndTime { get; set; }
    public string EventName { get; set; }
    public string EventDescription { get; set; }

    [TypeConverter(typeof(MultiNodeTreePickerConverter<Location>))]
    [UmbracoProperty("location")]
    public IEnumerable<Location> Locations { get; set; }
}
```

I would get a NotSupported exception if the user didn't select any locations in the back office. 

I fixed it by subclassing:

```
public class AllowEmptyMultiNodeTreePickerConverter<T> : MultiNodeTreePickerConverter<T> where T:class
{
    public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
    {
        if (value is string && value.Equals(""))
        {
            return Enumerable.Empty<T>();
        }

        return base.ConvertFrom(context, culture, value);
    }
}
```

And using that type as the TypeConverter in the attribute.

Looking at the code, `MultiNodeTreePickerConverter` returns an empty enumerable if passed a null to convert. 

Am I using this attribute incorrectly?

If not, I was wondering if you were interested in a PR that treated empty string the same?

Thanks for your time!

akatakritos:
Empty strings pass through to the base TypeConverter and for enumerable
types at least, this results in a NotSupportedException.

cc90af9 added support for accepting null values, and in pretty much
every case, empty string should be treated as null: no meaninful data
can be gleaned from it for conversion.

EnumConverter works without the check because the empty string gets
pased to the base ConvertFrom method which just returns the default enum
value. But we might as well be explicit about the expected return result
for empty strings.

Fixes #58

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/ConverterHelper.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/ConverterHelper.cs
@@ -84,5 +84,16 @@
 
             return ids.ToArray();
         }
+
+
+        /// <summary>
+        /// Gets a value indicating if the object is null or the empty string
+        /// </summary>
+        /// <param name="value"></param>
+        /// <returns>true if the value is null or string.Empty</returns>
+        public static bool IsNullOrEmptyString(object value)
+        {
+            return value == null || value as string == "";
+        }
     }
 }
--- a/src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/ContentPickerConverter.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/ContentPickerConverter.cs
@@ -48,7 +48,7 @@
         /// </returns>
         public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
         {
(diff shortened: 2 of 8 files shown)


JimBobSquarePants:
Merging this into my current branch will cause trouble if done automatically as I've deleted the ConverterHelper class in favour of a base TypeConverter. I'll merge things manually then close this. Thanks for the code! :+1:

akatakritos:
Cool. Glad to help!

PR #60, opened 2015-04-07
Sample of mocking IPublishedContent

mortenbock:


leekelleher:
Thanks @mortenbock - it's very much appreciated! #h5yr

@JimBobSquarePants - We can get started with our unit-tests now :metal:

JimBobSquarePants:
Whooooop! Excellent work @mortenbock

PR #62, opened 2015-04-08
Unit Tests: Basic Ditto Mapping

leekelleher:
I wanted to make a start on trying out some basic Ditto mapping in the unit-tests.

Does this look like the right approach? :yum:


leekelleher:
@JimBobSquarePants @mattbrailsford No worries if you haven't had chance... wondering whether to merge this in?

JimBobSquarePants:
I haven't had a chance to dig around. On the surface it certainly looks sane but I'm no expert. I say go for it. Any issues we can sort in one location then rather than chasing branches.

leekelleher:
Cool, merged into `develop` branch. As you say, we can sort it out as we progress.

Issue #36, opened 2015-03-04
Support for Relations?


leekelleher:
@Hendy [mentioned this in the Gitter chatroom](https://gitter.im/leekelleher/umbraco-ditto?at=54f6ec0bdf52f34314e52685).

Consider adding support for Relations.

    [RelationType("relationTypeAlias")]
    public object RelatedItems { get; set; }

Then Ditto could get the related node IDs for the associated `IPublishedContent` node and pass the value to the POCO property.

e.g. `@Model.RelatedItems` could return a `IEnumerable<IRelation>` object type?

Details of the [`IRelation` interface can be found here](https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Core/Models/IRelation.cs).

Then for more complex types a custom TypeConverter can be used:

    [RelationType("relationTypeAlias")]
    [TypeConverter(typeof(RelationsTypeConverter))]
    public IEnumerable<IPublishedContent> RelatedItems { get; set; }

Then the custom `RelationsTypeConverter` can work out how to convert an `IEnumerable<IRelation>` into `IEnumerable<IPublishedContent>`.

leekelleher:
One concern might be that the `IRelationService` call hits the database? Do we know if Umbraco caches the result in memory?

mattbrailsford:
I'd say make sure to use the Lazy<T> class so that they are only loaded when someone tries to access them. Don't want to slow down general Ditto for an edge case.

leekelleher:
:+1: on `Lazy<T>`.

At this stage, I'm only noting this as an idea.  If we think it is too much of an edge case, then we can shelve it.

JimBobSquarePants:
I've just pushed a new branch for you to look at. `feature/allow-relations`

This contains a change to the `PublishedContentContext` class.

The context `PropertyDescriptor` property now points to the property on the target class that is currently being converted. This allows us to parse any attributes on that property from within a custom `TypeConverter` Using the `PropertyDescriptor.Attributes` property.

I *think* that would be enough to allow what is needed. Am I correct?

leekelleher:
You're correct!

Actually thinking about it, this might remove the need for having my [`UmbracoDictionaryValue` attribute](https://github.com/leekelleher/umbraco-ditto/pull/22), as it could be handled the same way - within a specialised/custom TypeConverter.

Hendy:
@JimBobSquarePants perfect :)

[...]

Issue #18, opened 2015-01-20
Dictionary Items


leekelleher:
Consider the ability to map Umbraco dictionary items to POCO properties.

e.g.

    [UmbracoDictionaryItem("MyLabel")]
    public string SomeLabel { get; set; }

Also wondering whether `UmbracoPropertyAttribute` should support dictionary items (as a fallback), e.g. prefix with a hash `[UmbracoProperty("myPropertyLabelAlias", "#MyLabel")]`

leekelleher:
Closing issue; opened PR #22

PR #64, opened 2015-04-14
Umbraco Dictionary Value Converter

leekelleher:
Revisiting #18, #22 (and various comments in #36)

Combining a custom attribute with a custom converter, we can populate a model's property with a value from an Umbraco dictionary-item.

For example...

```csharp
[UmbracoDictionaryValue("MyLabel")]
[TypeConverter(typeof(DittoDictionaryValueConverter))]
public string SomeLabel { get; set; }
```

There are a couple of implementation decisions that I'd like to discuss with this PR...

* The `DittoConverterWithAttribute` base/abstract converter.
  * I did it this way, as I thought we could reuse it when we do a similar attribute/converter for Relations, (see #36)
  * I'm not 100% happy with the naming. It should have "Converter" suffix, would "DittoWithAttributeConverter" make sense? :persevere:

* I've ended up using `UmbracoHelper` within `DittoDictionaryValueConverter`, as the Umbraco core's `CultureDictionaryFactoryResolver` is marked as internal in v6.2.5 :weary: (I considered using `uQuery` static helper, but that's probably worse these days)

Open for comment and suggestions. Once we're happy with the approach, we can do similar for Relations.


mattbrailsford:
1) DittoAttributeDependantConverter?
2) Sounds fair to use UmbracoHelper when no other option. As long as it's self contained within the type converter it shouldn't be a problem.

One thing I am wondering is if there is anyway not to use 2 attributes. Now, that's not to say never use the UmbracoDictionaryValueAttribute, but given you are using DittoDictionaryValueConverter, shouldn't it's first action be to look up a dictionary item with a key of SomeLabel but if the UmbracoDictionaryValueAttribute is present, then use the key specified there? Not sure how this would affect the Relations datatype though.

leekelleher:
Thanks @mattbrailsford.

1. :+1: (I like it ... still open to suggestions)
2. Yeah, I'd prefer not to use it, but there aren't many alternatives for getting dictionary item values.

re: Using one attribute - I originally did this in PR #22. But it meant adding another specific check in the `.As()` extension method, which was starting to get a bit bloated.  We discuss it a bit more #36.

mattbrailsford:
Ok, I see what you mean. I'm just wondering though, could we have either a base class or an interface for "AlternativeValueProvider" attributes that have a single method "public object GetValue()" then grab the attribute based on the base class, thus we could have many alternative value provider attributes, but they all get handled in one claws?

JimBobSquarePants:
:+1: On falling back to a default label.

Relations are going to have a single string key also as far as I can figure from looking at some of @Hendy's code for a relations converter. Perhaps checking if such a key exists in both implementations overwriting `CanConvertFrom` and falling back to checking if an attribute exists in the base.

I dunno how expensive those calls are though.

mattbrailsford:
There you go, there is an overload for GetCustomAttribute that can search the ancestor hierarchy https://msdn.microsoft.com/en-us/library/4x2fs7yd%28v=vs.110%29.aspx

JimBobSquarePants:
I use the equivalent overload for the GetCustomAttributes here. https://github.com/leekelleher/umbraco-ditto/blob/develop/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs#L391

mattbrailsford:
My last comment was wrong, the boolean flag checks for attributes on ancester types which we don't want, but look at the docs for the standard GetCustomAttribute, it looks like the type check checks the base types anyways

https://msdn.microsoft.com/en-us/library/9zhcty09(v=vs.110).aspx

attributeType
Type: System.Type
The type, or a base type, of the custom attribute to search for.

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/DittoConverterWithAttribute.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            return base.CanConvertFrom(context, sourceType);
+        }
+
+        /// <summary>
+        /// Gets the attribute.
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <returns></returns>
+        public T GetAttribute(ITypeDescriptorContext context)
```
JimBobSquarePants: protected virtual :wink:

leekelleher: Doh, of course!

leekelleher:
@mattbrailsford I think I know what you mean, but it's not quite clicked yet. Are you suggesting that we have our own Ditto version of `TypeConverterAttribute`, so then we can inherit from and extend?

One problem is that `TypeConverterAttribute` is marked as sealed :rage4:

mattbrailsford:
No, I'm suggesting we have the UmbracoDictionaryValue attribute extend an
abstract class like AlternateValueAttribute which has a method GetValue
that must be implemented by the concrete class (in this case it would look
it up in the dictionary). We then add logic in .As() to see if any
attributes are defined that inherit AlternatValueAttribute and if so, call
the GetValue method and then use that as the property value.

What I'm suggesting then is we use the attributes to define alternative
value sources, and type converters to convert value types into something
more meaningful (if you think about it, UmbracoProperty could inherit the
base class too as its resolving a different value). We can do similar to
type converters and pass in a context.

I personally think this makes it real clean and separates things nicely.

Attribute = Resolve alternate value
TypeConverter = convert value to something else

Matt

leekelleher:
@mattbrailsford :tada: **click** :smile: Gotcha!  I likes it!

mattbrailsford:
ValueResolverAttribute ?

abstract object ResolveValue(DittoResolverContext context)

:)

leekelleher:
Since we'd want to prefix it with "Ditto", how about simplifying it to "DittoValueAttribute"?
With an abstract "GetValue" method? (with whatever context-y things we want to pass in)

leekelleher:
Closing PR.

The discussion has moved over to PR #65 (on-going at the time of writing), the `DittoValueAttribute` feels a better way forwards.

mattbrailsford:
IMO it's fine because it's limited to a single purpose, and it's all self contained. It is responsible for resolving an Umbraco property, no more, no less.

PR #65, opened 2015-04-15
Ditto Value Attribute

leekelleher:
Inspired by the discussion on #64, we can explore the idea of using attribute-based value providers.


leekelleher:
@mattbrailsford @JimBobSquarePants I've made a branch/PR for us to play with - feel free to add to it, etc.

mattbrailsford:
1) I wonder if we should pass in a context object, rather than a heap of individual parameters? (I can't say I have a reason for this though :))

2) I'm wondering if the Type parameter is needed? I reckon value attributes should just return raw values, so if the value is needed in another format, this should be handled by a type converter?

Matt

leekelleher:
1. Good point, "DittoValueContext"?
2. I agree. (I was passing through everything we had available from the `PublishedContentExtensions.GetTypedProperty` method ... ya know, just in case ;-))

mattbrailsford:
1) Sounds good to me :)
2) I guess it's arguable. Someone might say the Type is needed to be able to find the right value to return, but I'm not sure.

leekelleher:
2. I'm thinking that the `DittoValueAttribute` will get the raw value, then custom TypeConverters after that.

mattbrailsford:
To be honest, I'm wondering if you couldn't just reuse ITypeDescriptorContext (or at least create an interface on top of it (IValueDescriptorContext?)) as I think that has all you need access to. And it looks like Type isn't one of the parameters on the context, that is passed at the time of convert.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -315,54 +315,38 @@ public static class PublishedContentExtensions
...
-                                    ? contentProperty.GetValue(content, null)
-                                    : content.GetPropertyValue(umbracoPropertyName, recursive);
+            // The default return value.
+            object value = null;
 
-            // Try fetching the alt value.
-            if ((propertyValue == null || propertyValue.ToString().IsNullOrWhiteSpace())
-                && !string.IsNullOrWhiteSpace(altUmbracoPropertyName))
+            // Check if the property has an associated value attribute.
+            var valueAttr = propertyInfo.GetCustomAttribute<DittoValueAttribute>(true);
```
mattbrailsford: var valueAttr = propertyInfo.GetCustomAttribute<DittoValueAttribute>(true) ?? new UmbracoPropertyAttribute();

leekelleher: Good spot! #h5yr

leekelleher:
A question about `UmbracoPropertyAttribute`... what do you think about having code/logic within the attribute itself?

mattbrailsford:
IMO it's fine because it's limited to a single purpose, and it's all self contained. It is responsible for resolving an Umbraco property, no more, no less.

mattbrailsford:
The only thing I could think to do different is have the attribute reference a Type and it's that Type that is responsible for fetching the value. This seperates the fetching logic from the attribute itself, but would mean code is dotted around. I personally don't mind it being in the attribute.

leekelleher:
I'm happy with the code/logic being in the attribute itself. (I was only covering myself :wink:)

mattbrailsford:
Actually, I think I have an idea. Just out at the Mo, but will fill you in
when I'm back :)

mattbrailsford:
Hmm, didn't work out quite like I thought.

So my idea was to make the DittoValueAttribute more like the TypeConverterAttribute class in that it's only job is to reference a ValueResolver class that does all the hard work. This would mean DittoValueAttribute would never need to change as it would work just like the TypeConverterAttribute

    [DittoValueResovler(typeof(DictionaryValueResovler))]
    public string MyProp { get; set; }

This would mean the logic is held in simple resovlers and outside the attributes, and also follows the same pattern as the TypeConverters. Hurra!

Unfortunately I forgot to think about he constructor parameters, such as being able to supply an alternative alias field etc.

If we want to go this route then, we still could, but it would mean we have to create custom attributes along with our value resovler. The attribute's only job would be to collect setting values. Then the resolver could be passed the attribute on construction to retrieve the config values and do it's job of resolving the value.

We could make the DittoValueResovlerAttribute accept a params object[] typeParams collection so you can pass in any number of parameters to the attribute and pass them down such that you wouldn't need to make custom attributes, the only problem with that is that it's not very discoverable, ie, intellisense wouldn't tell you the parameters the resolver is expecting so you kind of have to magically know what's needed and in what order. Not great.

I do like the idea of making Attributes dumb, and using resolver classes to do the hardwork, it's just whether the extra work is worth it?

leekelleher:
> for ref, @mattbrailsford and I had been chatting on skype about this too

So we'd have something like this?

    [DittoValueResolver(typeof(UmbracoPropertyResolver), "alias", "altAlias", true, "default value")]

... then we could create attribute sugar for it using the existing `UmbracoPropertyAttribute`?

I guess we need to weigh up the benefits of taking this approach over having logic in the Attribute itself.

mattbrailsford:
So I've just pushed some code that separates resolver logic from attribute definition. Have a look and see what you think. 

There is a bit of room for improvement within the .As() extensions method, I've just added some TODO's for the time being.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -315,54 +315,33 @@ public static class PublishedContentExtensions
...
+            var valueAttr = propertyInfo.GetCustomAttribute<DittoValueResolverAttribute>(true)
+                ?? new UmbracoPropertyAttribute();
+
+            //TODO: Only create one context and share between GetRawValue and SetTypedValue?
+            var descriptor = TypeDescriptor.GetProperties(instance)[propertyInfo.Name];
+            var context = new PublishedContentContext(content, descriptor);
+
+            // Get the value from the custom attribute.
+            //TODO: Cache these?
+            var resolver = (DittoValueResolver)Activator.CreateInstance(valueAttr.ResolverType);
```
JimBobSquarePants: A wee heads up .We're using a single place for creating objects. An extension `type.GetInstance()` It's so we can keep the api consistent for creation as the overloads of that extension use much faster initialisation code than the default.

mattbrailsford: Swapped and pushed :)

JimBobSquarePants:
I'm really liking this addition. Very neat!

PR #67, opened 2015-05-14
Proxy

leekelleher:
@JimBobSquarePants I've opened a PR for the proxy branch, I know it doesn't currently merge automatically, but it gives us a place to iron out any issues.


JimBobSquarePants:
Excellent! I found it best to keep merging develop back into the proxy to make things easier. I need to have a good look at what has changed from your dictionary updates. As far as I'm aware the proxy bits are good to go. I've been using them for weeks now without issue.

JimBobSquarePants:
I've performed the merge. Will need to test to ensure all is hunky dory but it seemed to go smoothly enough.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -328,18 +387,18 @@ public static class PublishedContentExtensions
             Type type,
```
JimBobSquarePants: Type doesn't get used here?

src/Our.Umbraco.Ditto/Proxy/ProxyFactory.cs
```diff
@@ -0,0 +1,287 @@
...
+        /// The constructor arguments to pass to any proxy types.
+        /// </param>
+        /// <returns>
+        /// The proxy <see cref="Type"/> instance.
+        /// </returns>
+        public object CreateProxy(Type baseType, IInterceptor interceptor, IEnumerable<PropertyInfo> excludedProperties, params object[] constructorArguments)
+        {
+            Type proxyType = this.CreateProxyType(baseType, excludedProperties);
+
+            object result = Activator.CreateInstance(proxyType, constructorArguments);
```
JimBobSquarePants: Using `Activator.CreateInstance` as the function expression version couldn't determine the type. Could probably rewrite it so it knows what to expect.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
leekelleher: Cool, I'll remove it :+1:

src/Our.Umbraco.Ditto/Proxy/ProxyFactory.cs
leekelleher: OK, no worries.

Had a quick thought, maybe the `TypeInitializationExtensions.GetInstance` extension method?

JimBobSquarePants: That was the issue. `TypeInitializationExtensions.GetInstance` can't accept params, it can't determine the type properly. I'll probably create two separate method signatures since we know the two options.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -324,22 +383,21 @@ public static class PublishedContentExtensions
         /// <returns>The <see cref="object"/> representing the Umbraco value.</returns>
         /// <param name="instance">The instance to assign the value to.</param>
         private static object GetRawValue(
-           IPublishedContent content,
-            Type type,
+            IPublishedContent content,
             CultureInfo culture,
             PropertyInfo propertyInfo,
-            ref object instance)
+            object instance)
```
leekelleher: Just noticed that `ref` has been removed here. @mattbrailsford Does this have an impact?

mattbrailsford: In this method it shouldn't make a difference, but in the SetTypedValue method I'm not too sure. propertyInfo.SetValue gets called in that method meaning it modifies the instance object. If not passed in by ref then the calling code wouldn't get the updated changes. I'd have to look through the codebase to see what calls this method and whether it expects the instance object to be updated or not.

leekelleher:
I've tested the `feature/proxy` branch against my demo site on Carlsberg... and all good!

JimBobSquarePants:
It's clever stuff eh? :wink: 

Did you look at the difference in workload when using virtual properties?

I thought of something today. I need to add a serializable attribute to my proxy class to ensure it's cacheable. I can do that easy enough though.

leekelleher:
I didn't make any changes to the Carlsberg code/models (yet).  I first wanted to check that it worked as a drop-in DLL replacement, (which it does -- except for the "Ditto" prefixes breaking-change on our TypeConverters).

I'll do some benchmarks against the old version, then the same with latest proxy, then same with some virtual properties.

JimBobSquarePants:
I noted a big difference on the site I was working on, Miniprofiler really highlights the difference in activity.

leekelleher:
I've tried to have a play around with doing benchmarks this afternoon, but there's not much difference between 0.6.1 and 0.7.0 (with no model changes), maybe 100-ish ms.

With the virtual properties, I'm unsure how to test them within my Carlsberg demo site - as all the properties on my POCOs are being used/accessed in the views.  There's nothing that I need to have lazy loaded (yet).

@mattbrailsford Did you want to give this a test before we merge in to the develop branch?

mattbrailsford:
I'll see if I can't have a play over the w/e

JimBobSquarePants:
@leekelleher Do you have anything that lists descendants? Menu's, news section etc. That's where you'll see the biggest difference.

leekelleher:
@JimBobSquarePants I think we should merge this in now. We can do any polish in the `develop` branch prior to next release.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
leekelleher: @JimBobSquarePants @mattbrailsford Did we figure out what we need to do with the `ref` keyword?

JimBobSquarePants:
YAAAAAAAAAAAAAAAASSSSSSSSSSSSSSSSSS!!!!!!

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
JimBobSquarePants: I don't think it needs it. I'll double check.

Issue #72, opened 2015-05-27
web.config AppSetting resolver ?


Hendy:
How about a resolver to get a value from appSettings (in web.config) ?

leekelleher:
Good suggestion! It's totally possible... more of a question of do we want this in Ditto core, or side project? (e.g. [micklaw's Ditto.Resolvers](https://github.com/micklaw/Ditto.Resolvers))

Hendy:
well, no 3rd party references would required (and it'd be tiny)

Hendy:
it could, if the value was null, then use the property name as the key, or if a value had been set, then use that value for the key

mattbrailsford:
I'm not completely averse to this one. I think as long as it's pure Umbraco or pure .NET, then it could be core. Things like the Archetype one are 3rd party dependent, so a no no.

leekelleher:
:+1: Cool with me.  @Hendy do you feel up for a PR? :smile_cat:

JimBobSquarePants:
Nice!

Hendy:
Any prefix for AppSettingAttribute / AppSettingValueResolver ?

leekelleher:
Sounds good to me! :+1:

leekelleher:
Implemented in #73

PR #73, opened 2015-05-27
AppSetting Attribute & Resolver

Hendy:
as referenced by #72
(behaves in the same way as the UmbracoDictionary Attribute & Resolver)


leekelleher:
Thanks @Hendy! Just in time for the v0.7.0 release too! :smile:

Issue #74, opened 2015-05-28
make parameter optional for UmbracoDictionary and AppSetting resolvers


Hendy:
if a dictionary key, or an app setting key aren't provided with their corresponding attribute, then fall-back to using the property name

leekelleher:
:+1: 

You know the only reason I didn't do this for the Dictionary one is because on my website implementations the dictionary names are dot-namespaces, e.g. "Client.Project.Labels.Name".
So they'd never match up with the POCO property name.  But I can see the usage in this - so yes, totally valid issue!

Hendy:
Hi @leekelleher I should be able to do that later today

leekelleher:
Thanks @Hendy! #h5yr

leekelleher:
Closing issue, as PR opened #75

PR #75, opened 2015-05-28
Feature/resolver property name fallback

Hendy:
PR for issue #74

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/ValueResolvers/AppSettingValueResolver.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/ValueResolvers/AppSettingValueResolver.cs
@@ -28,12 +28,21 @@
         /// </returns>
         public override object ResolveValue(ITypeDescriptorContext context, AppSettingAttribute attribute, CultureInfo culture)
         {
-            if (string.IsNullOrWhiteSpace(attribute.AppSettingKey))
+            // default to attribute value
+            string appSettingKey = attribute.AppSettingKey;
+
+            if (string.IsNullOrWhiteSpace(appSettingKey) && context.PropertyDescriptor != null)
+            {            
+                // fall-back to property name
+                appSettingKey = context.PropertyDescriptor.Name;
+            }
+
+            if (string.IsNullOrWhiteSpace(appSettingKey))
             {
                 return null;
             }
 
-            return WebConfigurationManager.AppSettings[attribute.AppSettingKey];
+            return WebConfigurationManager.AppSettings[appSettingKey];
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Our.Umbraco.Ditto/ComponentModel/ValueResolvers/AppSettingValueResolver.cs
```diff
@@ -28,12 +28,21 @@ public class AppSettingValueResolver : DittoValueResolver<AppSettingAttribute>
         /// </returns>
         public override object ResolveValue(ITypeDescriptorContext context, AppSettingAttribute attribute, CultureInfo culture)
         {
-            if (string.IsNullOrWhiteSpace(attribute.AppSettingKey))
+            // default to attribute value
+            string appSettingKey = attribute.AppSettingKey;
```
leekelleher: I think `PropertyDescriptor` is never null, (but I could be wrong? @mattbrailsford?).

On another note, we could get this into a shorter statement, as we do in [`UmbracoPropertyValueResolver.cs` line 33](https://github.com/leekelleher/umbraco-ditto/blob/develop/src/Our.Umbraco.Ditto/ComponentModel/ValueResolvers/UmbracoPropertyValueResolver.cs#L33)

    var umbracoPropertyName = attribute.PropertyName ?? (context.PropertyDescriptor != null ? context.PropertyDescriptor.Name : string.Empty);

*(not that we're aiming for crazy single-lines-of-code)*

mattbrailsford: I can't think of any reason PropertyName would ever be null, so you could probably just use

`var key = attribute.PropertyName ?? context.PropertyDescriptor;`

Hendy: @leekelleher I thought the same about context.PropertyDescriptor never being null, but took the lead from the line you mention in UmbracoPropertyValueResolver

mattbrailsford: in all honesty, it doesn't hurt to be over defensive :)

leekelleher: Thanks @mattbrailsford @Hendy ... sure, null-reference checking is fine, (I'm usually an over-defensive coder :bowtie:)

@Hendy Did you want to amend the code? I'm happy to merge in whenever (we can pick up small tidy-ups later)

Hendy: happy to update it - is it safe to assume context.PropertyDescriptor.Name will never be null ?

leekelleher: Argh! Let's assume that it *might* be `null` :wink:  let's keep the check in.

leekelleher:
Hey @Hendy, I'm going to merge this in now... I'm aiming to get the 0.7.0 release out tomorrow, so gonna do some tidy up this evening.

PR #79, opened 2015-06-01
CurrentContent attribute and value-resolver

### LINKED ISSUE #78: Re-applying IPublishedContent to a POCO property
(Feature sparked from discussion in #77)

--

Let's explore the idea of having Ditto re-apply the `IPublishedContent` from a POCO on one of it's own properties.

An example with code would be...

```csharp
public class MyModel
{
    [DittoReapplyContent]
    public MyOtherModel MyProperty { get; set; }
}
```

Obviously `DittoReapplyContent` is an absolutely terrible name... *suggestion welcome!*

This would be the equivalent of doing: `foo.MyProperty = content.As<MyOtherModel>()`.

leekelleher:
(Closes #78)

--

Introduces new attribute and value-resolver to apply the current `IPublishedContent` to a POCO property.

An example...

```csharp
public class MyModel
{
    [CurrentContent]
    public MyOtherModel MyProperty { get; set; }
}
```

This would be the equivalent of: `foo.MyProperty = content.As<MyOtherModel>()`.


src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -545,6 +545,13 @@ public static class PublishedContentExtensions
                 // Simple types
                 result = propertyValue;
             }
+            else if (propertyValue is IPublishedContent)
+            {
+                // If the property value is an `IPublishedContent`, then we can use Ditto to map to the target type.
+                result = ((IPublishedContent)propertyValue).As(propertyInfo.PropertyType);
+
+                // TODO: [LK] Could this be also used to handle IEnumerable<IPublishedContent>?
+            }
```
mattbrailsford: Need to add additional check for IEnumerable<IPublishedContent> where to result type is IEnumerable<Entity> and do ditto cast there too

tests/Our.Umbraco.Ditto.Tests/PublishedContentTests.cs
```diff
@@ -55,6 +55,7 @@ public void Property_Returned()
         }
 
         [Test]
+        [ExpectedException(typeof(System.InvalidOperationException))]
```
leekelleher: @mattbrailsford I hit a snag here, where the `SimpleModel` has a string property and is assigned an `IPublishedContent` ... previously Ditto did a `ToString()` with it, (well, Umbraco's `TryConvertTo` did it) ... Now it throws an exception, (as it tried to construct a `string` object).

mattbrailsford:
Just to clarify, there are two things going on here.

The [CurrentContent] attribute sets the properties resolved value as the current contents IPublishedContent. The things that then converts the value into the end entity type (MyOtherModels in Lees example) is because a new clause has been added to the main ditto conversion code that if the raw value type is IPublishedContent and no better converter could be found for the resulting type, then it falls back to doing a Ditto .As() call.

A couple of things that play on my mind are, should the CurrentContent attribute be something more like [CurrentContentAs] and have that call .As() internally, such that it's all self contained and explicit, and doesn't rely on a "magic" fallback (one argument could be, just using CurrentContent could result in another type converter being found to be a better match and being used instead of .As() thus producing unexpected results?, so CurrentContentAs is explicit that .As() is used).

Maybe we have two attributes

    [CurrentContent]
    public IPublishedContent Node { get; set; }

    [CurrentContentAs]
    public MyModel MyProp { get; set; }

The other thing is that I can't decide if the magic .As() fallback in the ditto conversion code should be there? It does make sense, that if it can't find a better way to convert, and the type is IPublishedContent, then it does it's best to return the type you asked for, it's just you didn't explicitly tell it to use Ditto to resolve it.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
leekelleher: Is this as simple as checking for `propertyValue is IEnumerable<IPublishedContent>`? or do we have to get funky with all the `.IsGenericType` checks?

mattbrailsford: Not sure, I think might be able to do IsAssignableTo

JimBobSquarePants: `propertyValue.IsEnumerableOfType(typeof(IPublishedContent))` 

Honeymoon is next week :wink:

leekelleher: Thanks @JimBobSquarePants!

(See you next week - *does it sound weird that we'll see you on your honeymoon?*)

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -545,6 +545,13 @@ public static class PublishedContentExtensions
                 // Simple types
                 result = propertyValue;
             }
+            else if (propertyValue is IPublishedContent)
```
leekelleher: (from Skype) @mattbrailsford suggests that we change to:

```csharp
else if (propertyValue is IPublishedContent && propertyInfo.PropertyType.IsClass)
```

*making a note here so I don't forget later*

PR #83, opened 2015-06-08
Added code for UmbracoProperties attribute

mattbrailsford:
This adds a new `[UmbracoProperties]` attribute which can be applied at class level to make it easier to express all properties in the target model should look for Umbraco properties with a specific prefix. For example, you may have a model of 

    [UmbracoProperties(Prefix = "site")]
    public class Site
    {
        public string Name { get; set; }
        public string Description{ get; set; }
    }

And in this instance, Ditto will look for "siteName" and "siteDescription" Umbraco properties (with fall-backs to "Name" / "Description" if not found)


mattbrailsford:
If everyone is happy with this, we could also look at adding an [UmbracoDictionaryPrefix] attribute for the dictionary resolver (and maybe more).

JimBobSquarePants:
Why would you not name your properties in the Umbraco back office to match your POCO properties? 

Wouldn't you also want to do this in reverse? try looking for a property called `Description` then look for one called `siteDescription`.

mattbrailsford:
If you explicitly set a prefix, then the prefixed should take precedence. 

In the example I give, I create a site object to hold site settings so I can encapsulate site setting . In that instance, you wouldn't want model.site.siteName, you'd just want model.site.name, hence the attribute .

PR #87, opened 2015-06-14
Pass existing object instance to Ditto

leekelleher:
Added parameter to enable passing an existing instance of a typed object to `As<T>` .


tests/Our.Umbraco.Ditto.Tests/ExistingObjectInstanceTests.cs
```diff
@@ -0,0 +1,35 @@
...
+                MyProperty1 = value,
+                MyProperty2 = value
+            };
+
+            content.As<MyModel>(instance: model);
+
+            Assert.That(model.MyProperty1, Is.Not.EqualTo(value));
+            Assert.That(model.MyProperty2, Is.EqualTo(value));
+        }
+    }
```
mattbrailsford: Maybe add a test for the scenario where you have a type with a base type and you call .As only using the base type as the generic type param. In this situation, only the properties defined in the base type should have their properties overridden.

leekelleher: +1 will do.

(So far this is the only test that I wrote on the back of the "party bus"
from #CGRT15 LOL!)

PR #91, opened 2015-06-17
Looking at option to make cleaner value resolvers

mattbrailsford:
Currently when you create a value resolver, you have to override a ResolveValue function that has a bunch of params that get passed through, for example: 

capture1

What I've done in this pull request is added a bunch of props to the ValueResolver class to hold these params, and simplified the ResolveValue method to a parameterless function

capture2

An added advantage to this is that I've also added a Content property which casts the context.instance property to IPublishedContent as this is going to be a common action.

The only negative in my eyes is that it isn't entirely obvious what properties are available to you, though I think this could be resolved with some documentation. My imediate feeling though is that less code = better.

What do you guys think?


leekelleher:
I like this. Since the original `ResolveValue` is marked as `virtual`, we can still override it.

I agree with you that the properties are less discoverable, so yes, we need to document this stuff.
(Well, once it's all settled that is :shipit:)

Issue #94, opened 2015-06-18
Do we need the concept of ValueResovlerContexts?


mattbrailsford:
So in my DitFlo demo, I've been making use of ValueResovlers a lot to resovle view model properties. There comes a time though when the value depends on certain flags within the request. Now we could have the value resovler just access the request, or we can DittoIgnore the property and we construct the value ourselves within a controller, however I was wondering if the concept of a ValueResovlerContext, that can be set at runtime, could be a handy feature.

In the DitFlo demo we have a news listing page that lists news items in a paged manor. To get the news items I'm using a value resovler, but obviously, the value resolver needs to know the current page so that it can grab the right news items. 

This is my current implementation which directly accesses the HttpContext Request object  is as follows:

    public class NewsResolver : BaseNewsResolver
    {
        public override object ResolveValue()
        {
            var currentPage = long.Parse("0" + HttpContext.Current.Request["p"]);
            var pageSize = 2;

            var items = GetNews().ToList();
            var totalItems = items.Count;
            var totalPages = (long)Math.Ceiling(totalItems/(decimal)pageSize);

            currentPage = Math.Max(1, Math.Min(currentPage, totalPages));

            var pagedItems = items.Skip((int)(currentPage-1)*pageSize).Take(pageSize)
                .As<NewsPreviewLink>();

            return new PagedCollection<NewsPreviewLink>
            {
                CurrentPage = currentPage,
                PageSize = pageSize,
                TotalItems = totalItems,
                TotalPages = totalPages,
                Items = pagedItems
            };
        }
    }

However from discussions held here https://github.com/mattbrailsford/umbraco-ditflo/issues/1#issuecomment-113080492 I've come up with a possible solution using ValueResolverContexts that could make that as follows

**Controller**  

    public class NewsOverviewController : DitFloController {
        public ActionResult Index(RenderModel model, int p) {
            Ditto.RegisterResolverContext(new NewsResolverContext{
                CurrentPage = p
            };
            return CurrentView();
        }
    }

**Resolver**  

    public class NewsResolver : BaseNewsResolver
    {
        private NewsResolverContext _context;

        public NewsResolver()
            : this(new NewsResolverContext())
        { }

        public NewsResolver(NewsResolverContext ctx)
        {
            _context = ctx;
        }

        public override object ResolveValue()
        {
            var currentPage = _context.CurrentPage;
            var pageSize = 2;

            var items = GetNews().ToList();
            var totalItems = items.Count;
[...]

PR #95, opened 2015-06-18
Resolver Contexts

mattbrailsford:
Pull request for issue #94 

Currently needs review from @JimBobSquarePants as I'm currently having to instantiate a value resolver outside of GetInstance. See https://github.com/leekelleher/umbraco-ditto/pull/95/files#diff-81cfd4fcb771a0cad63163155354443cR416 for info.

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/PublishedContentContext.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/PublishedContentContext.cs
@@ -9,7 +9,7 @@
     /// The published content context for supplying contextual information about the 
     /// <see cref="IPublishedContent"/> that is undergoing conversion.
     /// </summary>
-    internal class PublishedContentContext : ITypeDescriptorContext
+    public abstract class PublishedContentContext : ITypeDescriptorContext
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="PublishedContentContext"/> class.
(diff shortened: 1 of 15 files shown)


JimBobSquarePants:
Apologies for the slow reply... Swamped.

>Currently needs review from @JimBobSquarePants as I'm currently having to instantiate a value resolver outside of GetInstance. See https://github.com/leekelleher/umbraco-ditto/pull/95/files#diff-81cfd4fcb771a0cad63163155354443cR416 for info.

I don't get what you mean by this? `type.GetInstance` is a replacement for `Activator.CreateInstance(type, params)` as far as I can see you are using `new` which is perfectly fine. Have I missed something?

I'm unsure about the use of classes that inherit but add no extra properties or methods. `DittoValueResolverContext` feels a bit like code smell doing that. 

There's a few StyleKitten :cat: issues I noticed in the commit and in the develop branch as a whole. It could use a tidy up in there to make it easier to see what is going on.

mattbrailsford:
Hey @JimBobSquarePants 

RE Activator usage, I actually refactored out the issue I mentioned so it's not really an issue anymore.

RE the DittoValueResolverContext, I would ordinarily agree, but IMO it makes the API much more explicit, ie, if I create a custom Ditto value resolver context, it's much more obvious to inherit DittoValueResolverContext than PublishedContentContext.

What are we using for code style checking? as nothing has officially been agreed so might be worth creating an official standard.

JimBobSquarePants:
Ah no worries... Thought I might be losing my marbles.

Ok. For cleaner API I'm up for that then. 

Throughout the code base I've been using standard StyleCop (spacing, namespace scope, use of "this" etc) minus explicit type declaration and copyright declaration rules.

I've also noted that a common pattern regarding explicit type declaration throughout the code base is as follows:

```
var localProperty = 1;
object otherLocalProperty = instance;
```
So object should always be explicitly declared.

Issue #88, opened 2015-06-14
"Map" versus "Convert"?


mattbrailsford:
In the context of the converting / converted events, does use of the word "convert" make sense, or should we use something more like "map"? or are these so similar it doesn't make much difference?

leekelleher:
Good question! I'll chew on it too.

I think we used "Convert" due to that we were using TypeConverters.

I guess with "Map", we're not strictly mapping... We're creating, mapping
and converting values/types.

Still good point to be discussed more.

JimBobSquarePants:
I originally thought stick with "Convert" but the more I think about it I would rename them to "Map". Conversion is simply sometimes the means by which we map the properties.

leekelleher:
Closing this issue, any discussion can continue on PR #96

PR #96, opened 2015-06-19
Rename "Convert" to "Map" 

mattbrailsford:
As mentioned in issue #88 I've gone through and renamed usages of "Convert" to a "Map" alternative.

Do all of these now make sense?

* DittoMapHandler
* DittoMapHandlerContext
* DittoOnMappingAttribute
* DittoOnMappedAttribute
*  etc

I wasn't sure if DittoMapHandler / Context should be DittoMappingHandler (though that might cause confusion with Mapping / Mapped)

Matt


leekelleher:
I'm not 100% sold on "Map" yet... I probably need to let it sink in.

Let's leave this open for a few days - if that's cool?

mattbrailsford:
For sure, that's why I did it as a PR as wasn't entirely happy myself. Doesn't slow as nicely does it.

mattbrailsford:
Only other "Map" option I can think of is the older style "OnBeforeMap" / "OnAfterMap"

leekelleher:
The --ing and --ed follow Umbraco Core's conventions, (the Before/After were in v4 APIs)

mattbrailsford:
Deciding against this

leekelleher:
Shall we delete the `feature/convert-to-map` branch? or keep for reference?

No biggie either way, just doing some housekeeping :neckbeard:

PR #99, opened 2015-06-26
Register Global Conversion Handlers

mattbrailsford:
During a discussion with @leekelleher he came up with a scenario that is not currently supported by the new conversion handlers which was previously supported by the events system which was having the ability to modify objects during conversion that are outside our control (ie, we can't add the conversion handler attributes to them).

Rather than reinstating the events system, what I have here is something similar to how TypeDescriptor works in that it allows you to register a handler for a type at runtime. The syntax for this is:

    Ditto.RegisterConversionHandler<MyObject, MyHandler>();

What this is saying is during the conversion of any object of type MyObject run the MyHandler conversion handler during it's conversion.

One thing @leekelleher and I were unsure of was the use of the Ditto prefix. It's intention is to be a facade to give access to non .As methods such that everything is accessed from a single location and you don't get into the thing of trying to remember where you register handlers (kinda the same reason we have everything sit direct on the Our.Umbraco.Ditto namespace, to keep things simple).

Another question kind of arrises from this in that should we also add the ability to register ValueResolvers globally? And possibly TypeConverters (this could just be a method that calls TypeDescriptor.AddAttributes(...) but might be nice to add it to Ditto facade just so everything can be accessed from one location.

Wadda you think?

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/ConversionHandlers/DittoConversionHandlerRegistry.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/ConversionHandlers/DittoConversionHandlerRegistry.cs
@@ -12,19 +12,17 @@ namespace Our.Umbraco.Ditto
         /// <summary>
         /// The cache for storing handler information.
         /// </summary>
-        private static readonly Dictionary<Type, IList<Type>> HandlerCache
-            = new  Dictionary<Type, IList<Type>>();
+        private static readonly Dictionary<Type, IList<Type>> Cache = new  Dictionary<Type, IList<Type>>();
 
         /// <summary>
-        /// The lock object to make HandlerCache access thread safe
+        /// The lock object to make Cache access thread safe
         /// </summary>
         private static object _cacheLock = new object();
 
         /// <summary>
         /// Static holder for singleton instance.
         /// </summary>
-        private static readonly Lazy<DittoConversionHandlerRegistry> _instance
-            = new Lazy<DittoConversionHandlerRegistry>(() => new DittoConversionHandlerRegistry());
+        private static readonly Lazy<DittoConversionHandlerRegistry> _instance = new Lazy<DittoConversionHandlerRegistry>(() => new DittoConversionHandlerRegistry());
 
         /// <summary>
         /// Private constructor to prevent direct instantiation..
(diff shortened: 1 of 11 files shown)


src/Our.Umbraco.Ditto/ComponentModel/ConversionHandlers/DittoConversionHandlerRegistry.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    /// <summary>
+    /// Registory for globaly registered conversion handlers.
+    /// </summary>
+    internal class DittoConversionHandlerRegistry
+    {
+        /// <summary>
+        /// The cache for storing handler information.
+        /// </summary>
+        private static readonly Dictionary<Type, IList<Type>> HandlerCache
```
JimBobSquarePants: Could probably use a ConcurrentDictonary here and remove the lock.

JimBobSquarePants:
:+1:   I think if you're gonna allow registration of one you might as well allow it for all. I actually do something similar for my querystring parser in ImageProcessor for TypeConverters (though without the generics) and it's damn useful.

https://github.com/JimBobSquarePants/ImageProcessor/blob/56dcc5b6225557d9ea27e06329ac0208752198f5/src/ImageProcessor.Web/Helpers/QuerystringParser/QueryParamParser.cs#L129

src/Our.Umbraco.Ditto/ComponentModel/ConversionHandlers/DittoConversionHandlerRegistry.cs
mattbrailsford: I can't because it's a list in a dictionary. With ConcurrentDictionary, access to the dictionary would be thread safe, but access to the list in the dictionary wouldn't, so I use a lock instead.

mattbrailsford:
Thanks for the feedback @JimBobSquarePants, much appreciated.

src/Our.Umbraco.Ditto/ComponentModel/ConversionHandlers/DittoConversionHandlerRegistry.cs
JimBobSquarePants: Ah right..

JimBobSquarePants:
No problemo :sunglasses:

PR #101, opened 2015-07-08
Enabled DittoConverter to pass back IPublishedContent 

robertjf:
Enabled DittoConverter to pass back IPublishedContent if the targetType is specified as such

Also fixed a typo on a private method...

Use Case:
I have a Content Model (Slide) that has an image and a target node as IPublishedContent Properties (because I don't know what type of node it is) and had to use the DittoMediaPickerConverter as not supplying a converter resulted in a null value coming back from the integer value that is the contentId.

This fix enabled the converters to pass back the resolved IPublishedContent successfully.

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/DittoConverter.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/DittoConverter.cs
@@ -9,7 +9,7 @@ namespace Our.Umbraco.Ditto
     using System;
     using System.ComponentModel;
     using System.Globalization;
-
+    using System.Collections.Generic;
     using global::Umbraco.Core;
     using global::Umbraco.Core.Models;
     using global::Umbraco.Web;
@@ -38,7 +38,13 @@ namespace Our.Umbraco.Ditto
                 return null;
             }
 
-            return UmbracoContext.Current.ContentCache.GetById(id).As(targetType, culture);
+            IPublishedContent content = UmbracoContext.Current.ContentCache.GetById(id);
+
+            // If we're expecting IPublishedContent then just return
+            if (targetType == typeof(IPublishedContent))
+                return content;
+
+            return content.As(targetType, culture);
         }
 
         /// <summary>
@@ -63,8 +69,17 @@ namespace Our.Umbraco.Ditto
             // Ensure we are actually returning a media file.
             if (media.HasProperty(Constants.Conventions.Media.File))
             {
+                // If we're expecting IPublishedContent then just return
+                if (targetType == typeof(IPublishedContent))
+                    return media;
+
                 return media.As(targetType, culture);
             }
+
+
+            // If we're expecting IPublishedContent then just return
+            if (targetType == typeof(IEnumerable<IPublishedContent>))
+                return media.Children();
 
             // It's most likely a folder, try its children.
             // This returns an IEnumerable<T>
--- a/src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/DittoConverter.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/DittoConverter.cs
@@ -78,7 +78,7 @@ namespace Our.Umbraco.Ditto
 
 
             // If we're expecting IPublishedContent then just return
-            if (targetType == typeof(IEnumerable<IPublishedContent>))
+            if (targetType.IsEnumerableOfType(typeof(IPublishedContent)))
                 return media.Children();
 
             // It's most likely a folder, try its children.
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -331,7 +331,7 @@
                                 () =>
                                 {
                                     // Get the value from Umbraco.
-                                    object propertyValue = GetResovledValue(content, culture, deferredPropertyInfo, localInstance, valueResolverContexts);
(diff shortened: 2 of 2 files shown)


leekelleher:
Thanks @robertjf - we'll review this.

On a side note, when I've hit the same use case, I've used `PublishedContentModel` as my property type.  (Since Ditto requires a concrete class, not an interface).

robertjf:
@leekelleher aha! hadn't considered that :)  yeah, that would work.

mattbrailsford:
Can you provide a unit test to show your use case?

robertjf:
er, yup... will work it up and add it to the PR

mattbrailsford:
Thanks, would just help clarify as I'm sure .As should already be able to return IPublishedContent if the prop type is IPublishedContent so this code shouldn't really be needed.

leekelleher:
Hmmm... I'd have thought that this line should have caught it:

https://github.com/leekelleher/umbraco-ditto/blob/develop/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs#L602

... but then would `IPublishedContent` pass the `IsInstanceOfType` check? I'm not sure, guess we need the unit-test.

robertjf:
I was getting Null returned on the ContentPicker and MediaPicker data types (stored value was id).

I then resorted to using the DittoContentPickerConverter which threw up errors:

        [TypeConverter(typeof(DittoContentPickerConverter))]
        public virtual IPublishedContent Target { get; set; }

which prompted me to put these checks in.  But I'll modify my code and trace through it again to check.

By the way, the latest NuGet package was through NullReferenceExceptions at me and I had to build and install from GitHub to overcome them.

leekelleher:
@robertjf - *another side note* - We've made a few changes (some breaking ones too :scream:) in the latest `develop` branch (which will be in upcoming v0.8.0 release).

robertjf:
ok, the problem with 
https://github.com/leekelleher/umbraco-ditto/blob/develop/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs#L602

is this:

* propertyInfo.PropertyType: {Name = "IPublishedContent" FullName = "Umbraco.Core.Models.IPublishedContent"}
* propertyValue: 1099 (it's an int)

so no, it's never going to catch it.

robertjf:
@leekelleher yep, I've taken note of them and modified my project code accordingly...

robertjf:
@mattbrailsford I'm not really sure how to go about setting up the data for the unit test, as the property converter will need to have a mock content store to create the the IPublishedContent from it's id - and then we need to test the actual TypeConverter - I suppose I could derive a `MockPropertyPickerTypeConverter` from the DittoConverter? But where to put it?

mattbrailsford:
OK, maybe we just change that to isAssingableFromType instead. Fire over
the unit test and we can try it out.

mattbrailsford:
Ahh, yes, fair point. Can you explain your exact setup then to reproduce?

robertjf:
sorry, had to step away.  So:
1. I have a DocType that has a standard ContentPicker (target) and a standard MediaPicker (image) on it (It's a slide, used by NestedContent but that shouldn't matter).
2. My Content Model looks like this:

    public class Slide : PublishedContentModel
    {
        public Slide(IPublishedContent content) : base(content) { }

        public virtual string Title { get; set; }

        [TypeConverter(typeof(DittoMediaPickerConverter))]
        public virtual IPublishedContent Image { get; set; }

        public virtual HtmlString Text { get; set; }

        [TypeConverter(typeof(DittoContentPickerConverter))]
        public virtual IPublishedContent Target { get; set; }
    }

Not using attributes ends up with null being returned;  the Picked content is coming through as an integer.

mattbrailsford:
@robertjf cool, thanks for that, will make it easier to test.

src/Our.Umbraco.Ditto/ComponentModel/TypeConverters/DittoConverter.cs
```diff
@@ -63,9 +69,18 @@ protected virtual object ConvertMediaFromInt(int id, Type targetType, CultureInf
...
+                // If we're expecting IPublishedContent then just return
+                if (targetType == typeof(IPublishedContent))
+                    return media;
+
                 return media.As(targetType, culture);
             }
 
+
+            // If we're expecting IPublishedContent then just return
+            if (targetType == typeof(IEnumerable<IPublishedContent>))
```
JimBobSquarePants: I'd like to use `IsEnumerableOfType(typeof(IPublishedContent)` here for testability and consistency.

leekelleher: :+1: Good spot!

JimBobSquarePants: It's mah thang! :sunglasses:

robertjf: yep, will update accordingly.

robertjf:
@JimBobSquarePants is that better?

JimBobSquarePants:
Aye... That's the one. Cheers! :+1:

mattbrailsford:
Ok, I've gone and pushed a better fix. Instead of having the converters doing the check, we now do it within .As. So where it would error before if it couldn't find a constructor for the type, as a final check before it errors it now checks to see if the content can be directly cast to the type. If it can, then it uses it, if not, then it goes on to throw the error as before.

robertjf:
cool, I'll test it out some time soon I'm sure :)

mattbrailsford:
We even have a lovely unit test :) https://github.com/leekelleher/umbraco-ditto/commit/668d804c0ea859396b994356a7a71f5f876849df#diff-e34ae8b20b17c7684fad59b31e36d5c0R302

PR #109, opened 2015-08-21
Rejected TypeConverter value returns null

leekelleher:
I found that if a TypeConverter's `CanConvertFrom` returns false, then the property's value is assigned a `null` reference.

I have added a fix, so that we check if the value type matches the target property's type, then we can assigned it.

It's probably easier to review the attached unit-test: [`ClassLevelTypeConverterTests`](https://github.com/leekelleher/umbraco-ditto/blob/hotfix-0.8.1/tests/Our.Umbraco.Ditto.Tests/ClassLevelTypeConverterTests.cs)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,5 @@
 # version format
-version: 0.8.0.{build}
+version: 0.8.1.{build}
 
 # UMBRACO_PACKAGE_PRERELEASE_SUFFIX if a rtm release build this should be blank, otherwise if empty will default to alpha
 # example UMBRACO_PACKAGE_PRERELEASE_SUFFIX=beta
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -587,6 +587,12 @@
                                     }
                                 }
                             }
+                            else if (propertyType.IsInstanceOfType(propertyValue))
+                            {
+                                // If the TypeConverter's `CanConvertFrom` has returned false,
+                                // then we can check if the value is the same type as the target type.
+                                result = propertyValue;
+                            }
                         }
                     }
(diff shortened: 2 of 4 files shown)


leekelleher:
Just waiting for the next round of @JimBobSquarePants's "ValueResolvers vs TypeConverters" battle :stuck_out_tongue_closed_eyes:

tweetle_beetles_b2_trimmed

JimBobSquarePants:
Cheeky!

The logic is definitely sound in this case since we want the user to get their property so it's a good fix. The question begs though, why would someone assign a TypeConverter to a property that it can't convert?

leekelleher:
It's a bit of an edge-case I have on a Carlsberg project.

I have a class, with a TypeConverter, which I use with regular `GetPropertyValue<T>` calls (outside of Ditto).
The property-editor is an Archetype, and the TypeConverter was originally written to only convert Archetype objects.

The complexity comes in as I need to have logic to recursively get Archetype values from up the tree (up to 3 levels) and aggregate (well, merge) the values.
(It's for an age-gate component, it gets complicated)

I'm using a ValueResolver to do this. (I bet you're already shaking your head? 😜)

So the ValueResolver is returning the target Type (of my custom class) ... but the TypeConverter didn't know what to do with it, so rejected it... and I got a `null` value.

I've fixed my TypeConverter to handle that scenario, but it got me thinking if other devs hit a similar issue.

Hope this makes sense?
(Sent from my mobile, excuse typos, etc)

JimBobSquarePants:
You kids and your ValueResolvers! :older_man: 

Certainly an edge case but yeah, since we already do the same thing under a different logic path 

https://github.com/leekelleher/umbraco-ditto/blob/a0ba43f56663794b3799edfec1051b509d8867b2/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs#L624

It's definitely wise to do it here also.

PR #115, opened 2015-09-17
Fixes DittoConversionHandlerRegistry cache type-cast

leekelleher:
The value type for `DittoConversionHandlerRegistry`'s internal `Cache` was being cast as an array, rather than a `List<Type>`.

This lead to a `NotSupportedException` ("Collection was of a fixed size.") when registering multiple conversion-handlers.

I've added a unit-test to support this fix.

@mattbrailsford I've assigned to you for a sanity check.


mattbrailsford:
Looks good to me

leekelleher:
Thanks @mattbrailsford!  I never thought I'd be using multiple `DittoConversionHandler` on a model-type.  Turns out that I hit an edge-case on Carlsberg, which I'm hoping to refactor (in due course), but still good to iron out this specific bug. :+1:

mattbrailsford:
Agreed

PR #118, opened 2015-09-24
Custom PublishedContent property nullifies value

leekelleher:
@Jeavon [raised an issue on Gitter](https://gitter.im/leekelleher/umbraco-ditto?at=5603f6cbe85e8d337252aa51) about Ditto not mapping the value of a property when using a custom `IPublishedContent` that has a property with the same name.

I have modified the `UmbracoPropertyValueResolver` to handle this scenario, (and included a supporting unit-test).

We first try to get the value from the `IPublishedContent` object property, then if the value is null, we make a call to the content's `GetPropertyValue` method.

---

@JimBobSquarePants @mattbrailsford Could either of you cast an eye over this change and scenario, (most as a sanity check), thanks.

@Jeavon If you want to try out a custom build from this branch to confirm that this works for you?

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/ValueResolvers/UmbracoPropertyValueResolver.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/ValueResolvers/UmbracoPropertyValueResolver.cs
@@ -59,9 +59,16 @@
             if (!umbracoPropertyName.IsNullOrWhiteSpace())
             {
                 var contentProperty = contentType.GetProperty(umbracoPropertyName, BindingFlags.Public | BindingFlags.IgnoreCase | BindingFlags.Instance | BindingFlags.Static);
-                propertyValue = contentProperty != null
-                    ? contentProperty.GetValue(content, null)
-                    : content.GetPropertyValue(umbracoPropertyName, recursive);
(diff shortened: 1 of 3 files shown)


Jeavon:
@leekelleher yup custom build works for us, thanks!

JimBobSquarePants:
Looks good to me! :shipit:

mattbrailsford:
If recursive is true, shouldn't it check all ancestor properties before checking Geypropertyvalue?

leekelleher:
@mattbrailsford Trying to get my head around your question... do you mean ancestor properties of the inherited `IPublishedContent` object? :confused:

mattbrailsford:
Yea. I'm not sure it's required actually, I just saw the recursive flag on
the second if statement and wondered if it should be recursive on the
first.

In theory, I'm not sure if a prop should be valueless, although I guess it
can be hence the need for this PR.

leekelleher:
I'll merge this in for a v0.8.3 patch release.

With the recursive question, I'm unsure what it means, but since this PR doesn't change how the values are retrieved, we can move the discussion to a future (0.9.0) release.

mattbrailsford:
Agreed

PR #126, opened 2015-10-14
Unit-test for ImageCropDataSet

### LINKED ISSUE #125: Unit-test for ImageCropDataSet
I keep seeing various conversations about having Image Cropper support in Ditto, (or similar).

Here's a recent discussion on the forum:
https://our.umbraco.org/projects/developer-tools/ditto/ditto-feedback/66353-imagecropper-mapping-with-ditto#comment-231497

I'm opening this ticket as a to-do for myself.  Write a unit-test that tries to map the Image Cropper data/value with the `ImageCropDataSet` object type.  More to satisfy my curiosity.

leekelleher:
Discussions around ImageCrop values/models keep arising, this unit-test is intended to clarify that the values can be mapped using Ditto.

Opening as PR, so to allow for any discussion. Do we need this unit-test in the Ditto core?

### DIFF
--- a/tests/Our.Umbraco.Ditto.Tests/ImageCropDataSetMappingTests.cs
+++ b/tests/Our.Umbraco.Ditto.Tests/ImageCropDataSetMappingTests.cs
@@ -1,0 +1,36 @@
+﻿namespace Our.Umbraco.Ditto.Tests
+{
+    using Newtonsoft.Json;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class ImageCropDataSetMappingTests
+    {
+        public class MyModel
+        {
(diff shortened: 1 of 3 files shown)


JimBobSquarePants:
Probably wise to cover it since the type is a core component. I've been mapping it in conjunction with a custom property value converter successfully for a while also. Remind me to share some code sometime.

leekelleher:
Since we started doing unit-tests on Ditto, whenever there's been an issue raised my mindset has been "let's unit-test it"!  Which is a good thing, just funny how it all came about.

JimBobSquarePants:
It's a really great approach. Something I am trying to adopt a lot more in other projects.

leekelleher:
Merged in. We can always remove it in future if we don't need it.

Issue #124, opened 2015-10-13
Issue of Custom Type Converter with newtwon Json


walleyuan:
I had an issue when using custom type converter to map a JObject with newtwon josn.  As newtwon json is using the type converter as well. so the custom type converter will be overwritten (find more details about the issue http://zhenyuan.azurewebsites.net/post/issue-of-custom-type-converter-with-ditto).  To walk around the issue, I just changed my serializer to javascriptserializer.

leekelleher:
Hi @walleyuan,

Thanks for raising this.  Although not really sure what Ditto can do about it though.

From your blog post, since the `LinkPicker` has the `LinkPickerTypeConverter` associated with it, so both Ditto and JSON.NET's (`ToObject`) would attempt to use that converter.

For the `JavaScriptSerializer` replacement, you could go with shorter syntax using: `JsonConvert.DeserializeObject<LinkPicker>(value.ToString())`

JimBobSquarePants:
I just got horribly burnt by this trying to serialize a class during processing against webapi. I have no idea how to work around it just now and I can imagine other developers are going to hit this issue.

leekelleher:
I believe JSON.NET has it's own version type-converter, `JsonConverter`.

Maybe we have a Ditto one too? e.g. `[DittoTypeConverter(typeof(MyCustomConverter))]`.  In which we use first, then fallback on the standard .NET TypeConverter attribute?  (I may need to explain this better?)

JimBobSquarePants:
I think we may need to. Come to think of that, I think AutoMapper does the same.

walleyuan:
I found the solution.  by dynamically adding the converter to your class. =P  @JimBobSquarePants

leekelleher:
@walleyuan Would it be possible to see the source of your `LinkPickerTypeConverter` class?

I'm trying to recreate the scenario in a unit-test, so we can figure out the best way to resolve it.

So far I've narrowed it down to overriding the `CanConvertTo` method to `return false;`.

// cc @JimBobSquarePants

leekelleher:
Closing this one.  I managed to recreate the issue with a unit-test and found that if my TypeConverter didn't implement `CanConvertTo`, then Newtonsoft.Json serialization would fail (null reference).

Feel free to re-open if we need further discussion.

PR #129, opened 2015-10-15
Switch out TypeConverter to allow serialization.

JimBobSquarePants:
If we attempt to serialize a class using JSON.NET that has a custom
TypeConverter bound the serialization will fail. This can have serious
implications when using ApiControllers. See #124 

This pull request switches out the `TypeConverter` base class and `TypeConverterAttribute` class for a custom `IDittoTypeConverter` and `DittoTypeConverterAttribute` types. 

Unfortunately it seems that we will not be able to maintain backwards compatibility. 

Answers on a postcard!

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoTypeConverterAttribute.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoTypeConverterAttribute.cs
@@ -1,0 +1,93 @@
+﻿namespace Our.Umbraco.Ditto
+{
+    using System;
+    using System.Diagnostics;
+    using System.Globalization;
+
+    /// <summary>
+    /// Specifies what type to use as a converter for the object this attribute is bound to. 
+    /// This class cannot be inherited.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.All)]
+    public sealed class DittoTypeConverterAttribute : Attribute
+    {
+        /// <summary>
+        /// Specifies the type to use as a converter for the object this attribute is bound to. 
+        /// This <see langword="static"/>field is read-only.
+        /// </summary>
+        public static readonly DittoTypeConverterAttribute Default = new DittoTypeConverterAttribute();
+
+        /// <summary>
+        /// The name of the type to use as a converter for the object this attribute is bound to.
+        /// </summary>
+        private readonly string typeName;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DittoTypeConverterAttribute"/> class.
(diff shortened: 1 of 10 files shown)


leekelleher:
@JimBobSquarePants I'll review this tomorrow, (as I'm at a meetup tonight). I have a couple of ideas that I'd like to try out.

JimBobSquarePants:
Great. Looking forward to seeing what  you can come up with.

JimBobSquarePants:
Looking at @walleyuan's comment and blog for issue #124 It looks like I can scrap all this. The following code allows full serialization/deserialization.

```
IEnumerable<Type> registerTypes = PluginManager.Current.ResolveTypes<MYBASECLASS>();

foreach (Type type in registerTypes)
{
    // Ensure that Json.NET can serialize/deserialize our types.
    TypeDescriptor.AddAttributes(type, new TypeConverterAttribute(typeof(JsonConverter)));
}
```

leekelleher:
I didn't get chance to look at this yet.

My understanding was that whenever someone (library or dev) wanted to find an associated TypeConverter for an object-type, they'd call `TypeDescriptor.GetConverter`, which would look for the `TypeConverterAttribute` (or if implemented `ICustomTypeDescriptor` - gets the details from there).

Since within Ditto we don't use `TypeDescriptor.GetConverter`, we check for the `TypeConverterAttribute` directly, ([see here](https://github.com/leekelleher/umbraco-ditto/blob/develop/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs#L499)), we could introduce our own `DittoTypeConverterAttribute`.  This would do exactly the same as `TypeConverterAttribute`, but standalone, e.g. .NET or other code libraries would be aware of it and try to associate the custom TypeConverter.

This way we wouldn't need to introduce our own `IDittoTypeConverter`, we'd still be using regular .NET TypeConverter classes, just having an additional way to register them with an object-type.

Does that make sense? Any flaws to the idea?

leekelleher:
@JimBobSquarePants The unit-test that you added `CanSerializeModelWithTypeAttribute`, I've tried adding it to the `develop` branch (locally) and it passes fine.

I'm wondering if we can come up with a unit-test that fails the JSON serialization?
Then we can work out a fix from that.

JimBobSquarePants:
Yeah, I'll have a go.

leekelleher:
Hey @JimBobSquarePants, how would you like to proceed on this one?

leekelleher:
@JimBobSquarePants I'm closing this one off.  I wont delete the branch yet, just in case you wanted to revisit it?

Issue #132, opened 2015-10-21
Base Class UmbracoPropertyAttributes ignore Classing Class UmbracoPropertiesAttribute


jamiepollock:
Hi all,
Lee's asked me to create an issue from [a topic](https://our.umbraco.org/projects/developer-tools/ditto/ditto-feedback/72324-umbracoproperties-prefix-reuse-with-a-similar-base-class) I had started on Our's Ditto forum.

Essentially if I had a base Ditto POCO with a few properties which are standard across except for the UmbracoProperties prefix of the calling class. 

Imagine having a sign up page which has content tabs for Sign Up Form & Sign Up Complete. Using an alt template Action name in the Controller to deviate content.

Base class:
```c#
    public class BasicContentViewModel : IBasicContentViewModel {
        [UmbracoProperty(AltPropertyAlias = "Name")]
        public string Title { get; set; }

        [UmbracoProperty]
        public HtmlString Content { get; set; }        

        [DittoIgnore]
        public bool HasContent { get { return Content != null && string.IsNullOrWhiteSpace(Content.ToString()) == false; } }
    }
```

Implementations using the prefix to change the content.
```c#
    [UmbracoProperties(Prefix = "signUpComplete_")]
    public class SignUpCompleteContentViewModel : BasicContentViewModel {
    }
```
Or...

```c#
    [UmbracoProperties(Prefix = "signUpForm_")]
    public class SignUpFormContentViewModel : BasicContentViewModel {
    }
```

Sadly as the `Title` or `Content` properties live in the `BasicContentViewModel` they don't pick up the `UmbracoPropertiesAttribute`. This is because of [these lines here](https://github.com/leekelleher/umbraco-ditto/blob/develop/src/Our.Umbraco.Ditto/ComponentModel/ValueResolvers/UmbracoPropertyValueResolver.cs#L30:L31).

`this.Context.PropertyDescriptor.ComponentType` resolves to BasicContentViewModel. It's a bit of a head scratcher because the PropertyDescriptor is core .net functionality.

Thanks,
Jamie

leekelleher:
Thanks @jamiepollock.  I've added a unit-test to support this in #133.

I've been looking at ways to get the correct `PropertyDescriptor.ComponentType`, but no joy so far.

jamiepollock:
That's what I concluded too from a generating a quick ValueResolver which returned `this.Context.PropertyDescriptor.ComponentType == null;` in `ResolveValue()`.

[...]

PR #133, opened 2015-10-21
Unit-test for inherited classes with prefixed properties

leekelleher:
Added a unit-test to verify the issue raised in #132 by @jamiepollock

The problem is that the Context in `UmbracoPropertyValueResolver` is not aware of the derived class.

It only knows about the `ComponentType`, which is the class that the property is bound to, (e.g. the inherited/base class).

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/PublishedContentContext.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/PublishedContentContext.cs
@@ -11,6 +11,11 @@
     public abstract class PublishedContentContext : ITypeDescriptorContext
     {
         /// <summary>
+        /// Gets the object <see cref="T:System.Type"/> of the instance that is being converted.
+        /// </summary>
+        public Type ConversionType { get; internal set; }
+
+        /// <summary>
         /// Gets the object that is connected with this type descriptor request.
         /// </summary>
         /// <returns>
--- a/src/Our.Umbraco.Ditto/ComponentModel/ValueResolvers/DittoValueResolverContext.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/ValueResolvers/DittoValueResolverContext.cs
@@ -1,10 +1,13 @@
 ﻿namespace Our.Umbraco.Ditto
 {
+    using System;
+
     /// <summary>
     /// The Ditto resolver context.
     /// Provides a way of providing contextual information to value resolvers.
     /// </summary>
     public class DittoValueResolverContext : PublishedContentContext
     {
+        public Type ConversionType { get; set; }
     }
 }
--- a/src/Our.Umbraco.Ditto/ComponentModel/ValueResolvers/DittoValueResolverContext.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/ValueResolvers/DittoValueResolverContext.cs
@@ -1,13 +1,10 @@
 ﻿namespace Our.Umbraco.Ditto
 {
-    using System;
-
     /// <summary>
     /// The Ditto resolver context.
     /// Provides a way of providing contextual information to value resolvers.
     /// </summary>
     public class DittoValueResolverContext : PublishedContentContext
     {
-        public Type ConversionType { get; set; }
     }
 }
--- a/src/Our.Umbraco.Ditto/ComponentModel/ValueResolvers/UmbracoPropertyValueResolver.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/ValueResolvers/UmbracoPropertyValueResolver.cs
@@ -25,9 +25,9 @@
             var altPropName = string.Empty;
 
             // Check for umbraco properties attribute on class
-            if (this.Context.PropertyDescriptor != null)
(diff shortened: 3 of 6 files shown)


leekelleher:
Wondering whether we merge this unit-test in? Obviously we'd change it to be aware of the failure.

mattbrailsford:
I think this could be possible, but would require some changes.

1) Pass the type passed into the .As call down to the ResolveValue function
2) Update the converter context to pass in the super type (would require an additional property on the PublishedContentContext which would mean deviating from the standard TypeDescriptor, or wrapping Type / Content into a container object and use that as Instance?)
3) Change the resolver code to use this new super type property

I'm not entirely sure the knock on effects though, but it is possible if we think it's worth pursuing?

mattbrailsford:
PS If we plan to create our own DittoTypeConverter base class, the custom property on PublishedContentContext wouldn't be a problem as we could define our type converters as using our custom context, and internally just cast it before use by the end ditto type converter.

leekelleher:
@mattbrailsford I'm still mentally processing this. :wink: 

It makes sense as currently our TypeConverters/ValueResolvers don't actually know anything about the actual Type that is being mapped, we make an assumption from checking `PropertyDescriptor` - which is only about the property (and its containing Type).

I'm okay with the idea of extending `PublishedContentContext` with having a "super type" property, (we'd need a more suitable name).  But I'm also wondering what the `IContainer` property is intended for? (_I haven't researched this yet._)

mattbrailsford:
yea, I had a look at that myself. I don't think IContainer is suitable really. Many we expose a TypeDescriptor property for the conversion Type?

leekelleher:
We can definitely explore this, but going back to @jamiepollock's original question (in #132), would you expect that scenario to work?

mattbrailsford:
Hard one isn't it. I guess whatever is defined on your master type should apply right? Then it's what should happen if you define an UmbracoProperties attribute on a base class and a super class?

leekelleher:
Going from @jamiepollock's example, I'd expect it to work.

As for what should happen if the `UmbracoProperties` attribute is defined on both base & super classes, I'm thinking that it would use whichever was passed into Ditto for conversation.

```csharp
[UmbracoProperties(Prefix = "base_")]
public class MyBaseModel { }

[UmbracoProperties(Prefix = "super_")]
public class MySuperModel : MyBaseModel { }
```

So when you call...

```
@Model.Content.As<MyBaseModel>()
```

It would use `UmbracoProperties` from `MyBaseModel`, and...

```
@Model.Content.As<MySuperModel >()
```

Would use `UmbracoProperties` from `MySuperModel` accordingly.

(Sorry for the verbose code snippets - I was working it out myself in code)

mattbrailsford:
Yea, I would tend to agree

leekelleher:
@mattbrailsford - I'm sure there'd be edge-cases for this, someone wanting to mix-n-match the prefixes, but that feels way out of scope.

@jamiepollock - any thoughts? (as in what you'd expect to work?)

jamiepollock:
@leekelleher the snippet you provided makes a lot of sense.

It's an awkward one because we're using PublishedContentContext and PropertyDescriptor it doesn't provide any way to find the actual class it just has its class the property originates from.

I think this is one of those things that should make sense from a C# inheritance point of view but sadly in actuality it doesn't work out in implementation, currently. ;)

leekelleher:
@jamiepollock Thanks for the feedback. :+1: As @mattbrailsford mentions above, we could extend `PublishedContentContext` to contain a reference to the "super type", so could get it to work as expected.  Now it's more of a question of - just because we can, does that mean we should? ;-)

jamiepollock:
That's good to know but you're right, should we?

For me, I felt like my case from the original issue was a valid one. The behaviour was unexpected but that's how just how C# PropertyDescriptor works, eh?

The questions is should that behaviour be supplemented to allow for this edge case?

jamiepollock:
Just read that last comment back and every paragraph ended with a question. Weird!

mattbrailsford:
Yes, that is how PropertyDescriptor works, but that fact is irrelevant really, as it's more a question of what we think ditto should do. To me, the more I think about it, the more I think it should use what is defined on the master type as you would expect attributes to trickle down via inheritance, and after all, we are explicitly saying what type we are working with, so we should be really using that, IMO.

leekelleher:
Cool, let's give it a go! :-)

(No rush though)

leekelleher:
Okay, okay, I said no rush, then I got curious, and I had 10 mins to spare... added code to this PR. All unit-tests pass :+1:

We may want to review the name of the "ConversionType" property? :wink:

mattbrailsford:
Hehe, cool. 

The only thing I'm wondering is if we need to add the custom type converter stuff so that they can have access to it too, not just VRs

leekelleher:
@mattbrailsford I was toying with it, we definitely can do. Last night was more to prove a point to myself.

leekelleher:
@mattbrailsford Updated PR with moving the `ConversionType` property to `PublishedContentContext`

PR #138, opened 2015-11-01
Testing IPublishedContent with IsEnumerableOfType extension method

leekelleher:
Hey @JimBobSquarePants, here's one for you :smirk_cat: 

Whilst experimenting with returning collections of a custom `IPublishedContent` class, Ditto wouldn't attempt to cast/map them any further.

I set up a quick unit-test to see if this was the case... (where `PublishedContentMock` is a custom implementation of `IPublishedContent`)

```
typeof(IEnumerable<Mocks.PublishedContentMock>).IsEnumerableOfType(typeof(IPublishedContent))
```

Should this be true or false?

---

I've added in a fix that swaps around the params for the `IsAssignableFrom` check.

### DIFF
--- a/src/Our.Umbraco.Ditto/Extensions/Internal/TypeInferenceExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/Internal/TypeInferenceExtensions.cs
@@ -22,7 +22,7 @@
         public static bool IsEnumerableOfType(this Type type, Type typeArgument)
         {
             Type t = type.TryGetElementType(typeof(IEnumerable<>));
-            return t != null && t.IsAssignableFrom(typeArgument);
+            return t != null && typeArgument.IsAssignableFrom(t);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


JimBobSquarePants:
Yeah, in my mind that should return true. I'll double check against the tests etc to make sure your fix doesn't have any unforseen side effects.

leekelleher:
Thanks @JimBobSquarePants!

This came about when I was reviewing a Merchello/Ditto workshop implementation. Merchello has an `IProductContent` which derives from `IPublishedContent`, which "should" play nicely with Ditto, but it wasn't. :blush:

The other unit-tests all pass, which either means the patch is cool... or we're missing test scenarios :wink:  (_gotta test 'em all!_)

leekelleher:
@JimBobSquarePants No pressure, just curious if you've had chance to look this one?

JimBobSquarePants:
Sorry... Yeah, looks good to me!

leekelleher:
Cool, merged in!

Issue #151, opened 2016-01-09
Swap Out Value Resolution Easily (e.g., for Polyglot)


Nicholas-Westby:
This is in regards to this thread: https://our.umbraco.org/projects/developer-tools/ditto/ditto-feedback/74048-ditto-with-polyglot

In short, I use [Polyglot](https://our.umbraco.org/projects/backoffice-extensions/polyglot/) to build websites with multiple languages. Here's what that looks like:

![polyglot](https://cloud.githubusercontent.com/assets/5933222/12218433/e68490f6-b6d3-11e5-80c7-4982533f007d.png)

In this case, I have an "About Us" content node of document type "Info Page". Under that is a translations folder. Under that are my translations. My default language is English, and that's stored in properties on the "About Us" node:

![english](https://cloud.githubusercontent.com/assets/5933222/12218444/201d7788-b6d4-11e5-805f-2914cd9adf6b.png)

Spanish is stored in the "About Us (Spanish)" node:

![spanish](https://cloud.githubusercontent.com/assets/5933222/12218449/406c7e30-b6d4-11e5-9360-78f34220d5b8.png)

Due to some URL rewrite magic, the URL for the English page is `/en-us/about-us` and the URL for the Spanish page is `/es-mx/about-us`. However, the translated pages are never routed to (i.e., only the parent node, "About Us" in this case, is routed to).

If I want "Body Text" in the current language (as indicated by the URL), I use this code:

```
var bodyText = Model.Content.LocalizedPropertyValue<string>("bodyText");
```

Notice that I'm using `LocalizedPropertyValue` rather than `GetPropertyValue`. That is an extension method I built that lives here (a custom library for my company): https://github.com/rhythmagency/rhythm.umbraco.extensions/blob/e7df4420dcb75e9216f67ef0cea915a4a867c36f/trunk/Rhythm.Extensions/Rhythm.Extensions/ExtensionMethods/PublishedContentExtensionMethods.cs#L157

What I'd like to do is use Ditto to handle all my mapping like this:

```
var page = Model.Content.As<InfoPage>();
var bodyText = page.BodyText;
```

That won't work though, as Ditto is calling `GetPropertyValue` rather than `LocalizedPropertyValue`. I could create a `ValueResolver`, but then I'd have to decorate each and every property on my model classes with an attribute. It seems like there should be a better way.

Here's one idea:

```
var config = MyHelper.GetInternationalConfig();
var page = Model.Content.As<InfoPage>(config);
[...]

PR #152, opened 2016-01-11
Default Processor Type

leekelleher:
Proposed solution to issue #151

Adds a way to register the default processor type, either globally or at class-level.

Includes unit-tests to support this feature.

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoDefaultProcessorAttribute.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoDefaultProcessorAttribute.cs
@@ -1,0 +1,29 @@
+﻿using System;
+
+namespace Our.Umbraco.Ditto
+{
+    /// <summary>
+    /// The Ditto default processor attribute.
+    /// Used for specifying that Ditto should use a default processor during object conversion.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    public class DittoDefaultProcessorAttribute : Attribute
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DittoDefaultProcessorAttribute"/> class.
+        /// </summary>
+        /// <param name="processorType">Type of the processor</param>
+        public DittoDefaultProcessorAttribute(Type processorType)
+        {
+            this.ProcessorType = processorType;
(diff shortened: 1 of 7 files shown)


LoganLehman:
Going back to #151 , if we don't have the ability to use the new Processor framework due to our dependencies (Ditto.Resolvers), is there a way to hijack Ditto's IPublishedContent source before it starts mapping?

leekelleher:
@LoganLehman You could hook in via the `onConverting` parameter/event.

(I'm not sure how that would workaround the "default resolver" issue?)

LoganLehman:
@leekelleher - I work with Nicholas Westby. I am coming at the problem he was experiencing a different way (just for testing purposes). Because we of how we are using Polyglot, translation nodes are kept in a translation folder underneath the content node. Polyglot automatically copies all of the properties from the node that the translation is made for it. 

My idea is to check if a translation node exists, and matches the segment in current url "/en-us/", "/es-mx/". When those conditions are met, it will actually change the IPublishedContent source from the original English node, and swap it for the appropriate translation node, and then have mapping continue as usual.

That should hopefully result in the translation content being assigned to our models instead of the original english node.

LoganLehman:
@leekelleher - I have tested this out. It seems as though I can reassign the `DittoConversionHandlerContext`'s `Content` property but it doesn't have an effect on the output in the models:

        [DittoOnConverting]
        internal void OnConverting(DittoConversionHandlerContext ctx)
        {
            ctx.Content = ctx.Content.GetLocalizedContent();
        }

leekelleher:
@LoganLehman Ah ok, it will be because it's updating the context's instance property.

How is Ditto being called initially? Using the `.As<T>()` call, or something else?

I'm wondering if you can call `GetLocalizedContent()` prior to calling Ditto?

LoganLehman:
@leekelleher It seems as though we are using .As<T>() for output of everything as far as I can tell. I am going to test something like this here a minute:

        public static T GetLocalizedContentAs<T>(this IPublishedContent currentNode) where T : class
        {
            return GetLocalizedContent(currentNode).As<T>();
        }

LoganLehman:
@leekelleher - What I did above seems to work. Thanks for helping me talk through it ;)

leekelleher:
@LoganLehman No problem at all, glad you got it working! :+1:

PR #161, opened 2016-04-13
Unit-test UmbracoPicker processor

leekelleher:
Following a discussion with @mattbrailsford earlier today, I started to play with mocking the UmbracoContext in order to test the UmbracoPicker processor.

So far I have manually mocked things like HttpContextBase and the PublishedContent/MediaCache objects - we may find that a mocking library (such as Moq) would be better, but I have limited knowledge on that... (totally open to advice on that!)

The interesting part about testing UmbracoPicker is that we can remove the subsequent `.As(targetType)` calls, as when a processor returns an `IPublishedContent` object, it will automagically check the target type and process further.

---

I've opened as a pull-request to gauge feedback on the mocking approach... and to check that @JimBobSquarePants is happy with the `.As(targetType)` removal :smile:

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/Processors/UmbracoPickerAttribute.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/Processors/UmbracoPickerAttribute.cs
@@ -24,27 +24,16 @@ namespace Our.Umbraco.Ditto
         public override object ProcessValue()
         {
             // ReSharper disable once ConditionIsAlwaysTrueOrFalse
-            if (Context == null || Context.PropertyDescriptor == null)
+            if (Context == null || Context.PropertyDescriptor == null || Value.IsNullOrEmptyString())
             {
                 return Enumerable.Empty<object>();
-            }
-
-            var propertyType = Context.PropertyDescriptor.PropertyType;
-            var isGenericType = propertyType.IsGenericType;
-            var targetType = isGenericType
-                                ? propertyType.GenericTypeArguments.First()
-                                : propertyType;
-
-            if (Value.IsNullOrEmptyString())
-            {
-                return EnumerableInvocations.Empty(targetType);
             }
 
             // Single IPublishedContent
             IPublishedContent content = Value as IPublishedContent;
             if (content != null)
             {
-                return content.As(targetType, Context.Culture);
+                return content;
             }
 
             // ReSharper disable once PossibleNullReferenceException
@@ -53,7 +42,7 @@ namespace Our.Umbraco.Ditto
             // Multiple IPublishedContent
             if (type.IsEnumerableOfType(typeof(IPublishedContent)))
             {
-                return ((IEnumerable<IPublishedContent>)Value).As(targetType, Context.Culture);
+                return ((IEnumerable<IPublishedContent>)Value);
             }
 
             int[] nodeIds = { };
@@ -113,7 +102,7 @@ namespace Our.Umbraco.Ditto
                     }
                 }
 
-                return multiPicker.As(targetType, Context.Culture);
+                return multiPicker;
             }
 
             return null;
--- a/tests/Our.Umbraco.Ditto.Tests/Mocks/MockHttpContext.cs
+++ b/tests/Our.Umbraco.Ditto.Tests/Mocks/MockHttpContext.cs
@@ -1,0 +1,9 @@
+﻿using System.Web;
+
+namespace Our.Umbraco.Ditto.Tests.Mocks
+{
+    public class MockHttpContext : HttpContextBase
+    {
+        public override HttpRequestBase Request { get { return null; } }
+    }
+}
--- a/tests/Our.Umbraco.Ditto.Tests/Mocks/MockPublishedContentCache.cs
+++ b/tests/Our.Umbraco.Ditto.Tests/Mocks/MockPublishedContentCache.cs
@@ -1,0 +1,76 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Xml.XPath;
+using Umbraco.Core.Models;
+using Umbraco.Core.Xml;
+using Umbraco.Web;
+using Umbraco.Web.PublishedCache;
+
+namespace Our.Umbraco.Ditto.Tests.Mocks
+{
+    public class MockPublishedContentCache : IPublishedContentCache
+    {
+        public bool XPathNavigatorIsNavigable
+        {
+            get
+            {
+                throw new NotImplementedException();
+            }
+        }
+
+        public IPublishedContent CreateFragment(string contentTypeAlias, IDictionary<string, object> dataValues, bool isPreviewing, bool managed)
+        {
+            throw new NotImplementedException();
+        }
+
+        public IEnumerable<IPublishedContent> GetAtRoot(UmbracoContext umbracoContext, bool preview)
+        {
+            throw new NotImplementedException();
+        }
+
+        public IPublishedContent GetById(UmbracoContext umbracoContext, bool preview, int contentId)
+        {
+            return new PublishedContentMock { Id = contentId };
+        }
+
+        public IPublishedContent GetByRoute(UmbracoContext umbracoContext, bool preview, string route, bool? hideTopLevelNode = default(bool?))
+        {
+            throw new NotImplementedException();
+        }
+
+        public IEnumerable<IPublishedContent> GetByXPath(UmbracoContext umbracoContext, bool preview, XPathExpression xpath, XPathVariable[] vars)
+        {
+            throw new NotImplementedException();
+        }
+
+        public IEnumerable<IPublishedContent> GetByXPath(UmbracoContext umbracoContext, bool preview, string xpath, XPathVariable[] vars)
+        {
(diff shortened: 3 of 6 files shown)


JimBobSquarePants:
STOP THE BUS! It what whats now?

I'm assuming that somewhere down the line our methodology looks at the property and converts it?

Does that happen only for known types with a property value converter i.e. `ImageCropDataSet` or does it work for custom models without converters. Something like this.

```
    /// <summary>
    /// Represents a single published image in the media section.
    /// </summary>
    [TypeConverter(typeof(DittoPickerConverter))]
    public class Image : IEntity, IFile
    {
        /// <inheritdoc/>
        public virtual int Id { get; set; }

        /// <inheritdoc/>
        [DeependSearchMergedField("nodeName")]
        public virtual string Name { get; set; }

        /// <inheritdoc/>
        public virtual string DocumentTypeAlias { get; set; }

        /// <inheritdoc/>
        public virtual int Level { get; set; }

        /// <inheritdoc/>
        public virtual DateTime CreateDate { get; set; }

        /// <inheritdoc/>
        public virtual DateTime UpdateDate { get; set; }

        /// <inheritdoc/>
        [UmbracoProperty(Constants.Conventions.Media.File)]
        public virtual string FileName { get; set; }

        /// <inheritdoc/>
        [UmbracoProperty(Constants.Conventions.Media.Bytes)]
        public virtual int Bytes { get; set; }

        /// <inheritdoc/>
        [UmbracoProperty(Constants.Conventions.Media.Extension)]
        public virtual string Extension { get; set; }

        /// <summary>
        /// Gets or sets the width in pixels.
        /// </summary>
        [UmbracoProperty(Constants.Conventions.Media.Width)]
        public virtual int Width { get; set; }

        /// <summary>
        /// Gets or sets the height in pixels.
        /// </summary>
        [UmbracoProperty(Constants.Conventions.Media.Height)]
        public virtual int Height { get; set; }

        /// <summary>
        /// Gets or sets the crops.
        /// </summary>
        [EditorBrowsable(EditorBrowsableState.Never)]
        [UmbracoProperty(Constants.Conventions.Media.File)]
        public virtual ImageCropDataSet Crops { get; set; }
    } 
```

leekelleher:
:warning::hand::bus:  Yup, let's defo pause and review :+1: 

With the processors feature, we have several default "post-processors", one of which - [`RecursiveDittoAttribute`](https://github.com/leekelleher/umbraco-ditto/blob/develop/src/Our.Umbraco.Ditto/ComponentModel/Processors/RecursiveDittoAttribute.cs#L23-L37) - handles the property-type check and can pass it back to Ditto for further mapping.

I'm not sure if we have a unit-test for using attributes on the target type - I'll double check.

leekelleher:
Just to note, that we still have the [`TryConvertToAttribute`](https://github.com/leekelleher/umbraco-ditto/blob/develop/src/Our.Umbraco.Ditto/ComponentModel/Processors/TryConvertToAttribute.cs) processor run at the end, so any custom types with custom TypeConverters will be called, (but that's from Umbraco's `TryConvertTo` rather than Ditto).

leekelleher:
Hmm, I can't find anything about having a custom processor set on the target type. Thinking more about it now, I'm not sure that Ditto handled this scenario previously, it was a side-effect of Umbraco's `TryConvertTo` calling the target type's TypeConverter attribute.

(This will still work btw - it just wont use the new processors)

@mattbrailsford What is your take on this? Should we be able to set a Ditto processor on the target object type?

mattbrailsford:
Hmm, we could make it work. I guess it would just be down to the processor
to check the raw value and see if its anything it can work with.

Don't we allow registering global processors which is effectively this
already, you just don't define them on the type.

JimBobSquarePants:
Just to note. I am using that exact class referenced in my code right now with the current. 0.83 release. 

I would intend to switch out the typeconverter for a processor with 0.9.0. Being able to set the processor on the object type is a feature we definitely should not lose. The original behaviour was that setting the converter was possible on the class but overridable on subsequent properties. 

On a side note, do we have a lot more code than we used to? I've not looked at the commits but it does appear to me that everything seems to be a lot more complex.

JimBobSquarePants:
Check PublishedContentExstensions.cs line 499 in the master branch for original behaviour. (Can't copy paste from mobile for some reason. )

mattbrailsford:
I can't see it being a big deal adding it back in if it's missing (I'm not
even sure if it is). Out tomorrow morning but I'll take a look in the
afternoon.

Re added complexity, I personally don't think it's any more complex than
before, in fact we have fewer concepts and imo the code has been made way
more modular. The PublishedContentExtension files job is now just to
iterate properties with all the heavy lifting moved into processors.

It may have added a bit more code and some base classes, but that'll only
be to define the processor logic (and caching is in there now too).

Have a dig, I would hope you'd think it's a lot more logical than it was
before.

mattbrailsford:
Ok, so it is actually in there already that we test to target type aswell for any processor attributes

https://github.com/leekelleher/umbraco-ditto/blob/develop/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs#L473

I think they only question in this area (which I think Lee and I discussed at the time, though not sure if we came up with a definitive answer) is what to do if you have a type registered processor, but you also have processors defined on the property of your model? Right now it checks the property, and if non exist on the property, it defaults to the umbraco property processor, then it checks the type and appends that onto the list of processors, then it checks for globally registered processors and appends those, and fhen it finally appends the core processors for things like type converters and the recursive / enumerable processors and then run's through each of these one by one.

I guess the question is, should it be doing this "appending"? or should property defined processors override type defined processors?

leekelleher:
@mattbrailsford Oh cool, I couldn't find the corresponding unit-test for it.  (There's a default processor one).  But great that we've already got it covered :+1: 

In my mind Processors are a pipeline, rather than an either/or.  Question is when should the object-type's attribute be placed in the queue? (I'd go with after the property's attributes).

JimBobSquarePants:
@mattbrailsford  @leekelleher I guess I'm still struggling with the massive amount of changes. Takes a lot to get my head round - I can't really tell from the descriptions on each processor what it's use case is.

I think we need to make some good notes about the pipeline. That we are having to check the code for things like this suggests that no-one of us truly has a full understanding of the codebase any more. 

Can one of you talk me through how processors are collected and their processing order determined? It looks like there is an `Order` property but I'm not sure how the developer will know what number to set.

- What stops iterating through the list? It looks like each processor will attempt to process the value.
- What determines a `PostProcessorAttribute`? (`RecursiveDittoAttribute` is one and I don't know why)
- Do I now have to place multiple attributes on a property or class when one before would do?

Lots of questions I'm afraid...

mattbrailsford:
> That we are having to check the code for things like this suggests that no-one of us truly has a full understanding of the codebase any more.

Not exactly true, stuff slips peoples minds, I'm sure you don't recall every inch of IP all the time. The point is, I knew where to look to check.

Overall I think @leekelleher and I have quite a good understanding of how Ditto runs, and in reality, the only bit we don't understand is the lazy stuff. Which we don't see the problem in not knowing everything as long as we have someone on the team that does :)

In answer to your questions:

- Nothing stops processing through the list, each processor will be asked to process the value from the previous processors (or IPublishedContent for the first one). It's up to the processor to analyse the value / context information to see if it should run or not, if not, it should just return the value back out to pass onto the next processor.
- PostProcessorAttributes are currently all the post conversion stuff we did previously in one messy function broken out into individual processors, such as looking for enums and enumerating them, or casting from enum to individual instances / vica versa, or if a property returns IPublishedContent then recursively convert it (this is what the RecursiveDittoAttribute is). These are currently a hard codes list as per what Ditto used to do, but there is potential to make this publicly alterable so that people can easily add processors they want to run on every conversion.
- No. If you could do it in one attribute before, you can do it in one attribute now. There was just a lot fo confusion between TypeConverters and ValueResovlers previously, and there were scenarios where you'd use both and sometimes needed more, so this is why Processors come in to play. The do the job of both of these things, but you can chain them, so you can have much more granular control over what one of them does, ie, rather than having a ValueResolver, that fetches this, and converts that, and changes it's type into this, you can "if you want" separate these into reusable processors that you can chain together. This is where the Order property on the attribute comes in, as you can't rely on the order of attributes from the GetCustomAttributes call, so we have the order property so that you can define the order manually. Not perfect, but the best that can be achieved.

JimBobSquarePants:
@mattbrailsford Cripes you must type fast! :smile: 

That makes a lot more sense to me now thank you!

What I think I need to do now really is some dogfooding and convert something using a nightly. I'm certain that when I start using it I will immediately see the benefits in the changes you've made.

mattbrailsford:
@JimBobSquarePants no worries.

I think dog fooding really helps as this is where all of this stuff has come from, and finding where the weak points are so nothing has been added for the sake of it, it's all with a view to streamlining, but I can appreciate it's all been a bit fast paced and some quite broad changes, so it is tricky keeping up if you aren't following closely.

Docs is also something sorely needed, it's just sitting down and doing it (as is always the case)

leekelleher:
OK cool, back to the original PR - all happy for it to be merged in?

PR #162, opened 2016-04-13
UmbracoDictionary processor - added factory

leekelleher:
Following a discussion with @mattbrailsford earlier, (yup same chat that sparked PR #161), we agreed that using a ProcessorContext to abstract Umbraco logic wasn't the correct approach.
However the reason I used it was so I could unit-test the `UmbracoDictionary` processor.

Matt suggested that I try using a factory pattern to set the dictionary retrieval method. So this PR is my attempt.

Currently I've placed all the internal classes within the "UmbracoDictionaryAttribute.cs", I'm not 100% happy with this, but I wasn't sure where else to put them? (I'm open to suggestions)


mattbrailsford:
Hope you don't mind, but after giving this some more thought, I don't think the factory pattern is the right solution either, as a factory generally creates multiple types on demand, where as we were creating one type, and storing it for retrieval at any point, which isn't a true factory.

After searching around I found a nice solution here https://lostechies.com/jimmybogard/2008/11/09/systemtime-versus-isystemclock-dependencies-revisited/ where they use a replaceable delegate to fetch a value, but they allow the fetching logic to be replaced so that it can be easily unit tested.

I've updated the code accordingly and I think it makes it a bit cleaner. Wadda you think?

leekelleher:
@mattbrailsford Ace, I like it!

JimBobSquarePants:
Looks good. Will need more time to get my head around everything still though.

Issue #163, opened 2016-04-15
Question: Merge Ditflo with Ditto?


mattbrailsford:
Given the small size of Ditflo and the potential confusion people not knowing what it is compared to Ditto, I wonder if we should just merge them together and it's all just Ditto? You can of course still use ditto without using the Ditflo stuff (obviously this would be renamed to just be Ditto on merge), it just puts it all in one library so there is no need to explain the difference.

leekelleher:
:100: from me.

I see the `DitFloView` directive as complimentary to the `.As<T>` extension method. It is another entry point for Ditto.

mattbrailsford:
Closing, see PR #164 instead

PR #164, opened 2016-04-15
Merge in Ditflo

mattbrailsford:
As suggested in Issue #163, this is a version of Ditto with Ditflo merged in and all names changed from DitFloXxx to DittoXxx.

Important to note is in this PR, the baseline version of Umbraco has been upgraded to v 7.3.2, this is because Ditflo defines a base controller which makes it susceptible to this issue http://issues.umbraco.org/issue/U4-7399 This issue was fixed in 7.3.2 with the creation of the IRenderController which is what it is currently doing. If we want to keep Ditto supporting older versions, then we can by having it implement IRenderMvcController but note that by doing so makes it susceptible to the aforementioned issue.

Discuss! :)


src/Our.Umbraco.Ditto/App_Data/NuGetBackup/20160415-151441/CopyLogs/UmbracoBinBackup.log
```diff
@@ -0,0 +1,82 @@
+
```
leekelleher: This file can be deleted. It's from the NuGet package update. We can add an ignore rule for future reference.

src/Our.Umbraco.Ditto/Our.Umbraco.Ditto.csproj
```diff
@@ -33,52 +33,193 @@
...
+    </Reference>
+    <Reference Include="Owin">
+      <HintPath>..\packages\Owin.1.0\lib\net40\Owin.dll</HintPath>
+    </Reference>
+    <Reference Include="Semver">
+      <HintPath>..\packages\semver.1.1.2\lib\net45\Semver.dll</HintPath>
+    </Reference>
+    <Reference Include="SQLCE4Umbraco">
+      <HintPath>..\packages\UmbracoCms.Core.7.3.2\lib\SQLCE4Umbraco.dll</HintPath>
+    </Reference>
```
leekelleher: All these assembly references get added back in when a NuGet upgrade happens. It's a pain, not sure how to workaround it? I'm happy to tidy it up prior to merging in.

src/Our.Umbraco.Ditto/Our.Umbraco.Ditto.csproj
```diff
@@ -33,52 +33,193 @@
...
+    <Reference Include="Umbraco.Web.UI">
+      <HintPath>..\packages\UmbracoCms.Core.7.3.2\lib\Umbraco.Web.UI.dll</HintPath>
+    </Reference>
+    <Reference Include="UmbracoExamine">
+      <HintPath>..\packages\UmbracoCms.Core.7.3.2\lib\UmbracoExamine.dll</HintPath>
+    </Reference>
+    <Reference Include="UrlRewritingNet.UrlRewriter">
+      <HintPath>..\packages\UmbracoCms.Core.7.3.2\lib\UrlRewritingNet.UrlRewriter.dll</HintPath>
+    </Reference>
+    <Reference Include="WebMatrix.Data">
```
leekelleher: Where the heck does WebMatrix assemblies come from? Assuming it's some nested-dependency of UmbracoCms.Core.  No action needed, just commenting :open_mouth:

leekelleher:
Code-wise, looks good to me :+1:   (but I'm always shocked at how much "meta-dependency-crap" that Umbraco's NuGet adds into a project ... oh what we'd give for an "Umbraco.Core" only NuGet dependency! :heart_eyes:)

re: Umbraco version... I'm not precious about v6.x support, (selfishly, since I haven't used v6 in 2 years). Not sure if @JimBobSquarePants still uses it?  It would be great to have some rough stats on this.
However I am still using v7.2.8 on various Carlsberg projects, so it's whether I'd consider upgrading them to Ditto vNext or not? (I'll give it some thought)

mattbrailsford:
I guess one other option could be to put all this stuff in a Ditto.Web.dll which targets 7.3.2 whilst the main Ditto.dll remains on 6.x? Can still form part of the same package install, it's just spread into two DLL's so that they can target different baselines?

mattbrailsford:
Although, I guess you are still in a position where it won't work right for people prior to 7.3.2 if they try to use any of the web stuff, so unless you were to create 2 packages (which I'd want to avoid) I guess it's still either upgrade, or introduce that bug

leekelleher:
@mattbrailsford Do you have any urgency to have this merged in? I'm thinking that we release 0.9.0, as the docs are ready... and we merge this in for a 0.10.0 release. (_Pending Umbraco version discussion._) We don't have to wait too long between releases, also it would be good to have docs ready for Ditflo.

mattbrailsford:
@leekelleher nope not really, other than my presentation next week (27th April) which might have been nice to just do it on ditto, rather than ditto + ditflo and then it instantly changes. Though, I guess if we are committing to merge it in, I can still do it as Ditto, and just says it's coming in the next release?

leekelleher:
@mattbrailsford Go with calling it all "Ditto", I'm sure you'd mention Ditflo and the history of the projects anyway, so it should all pan out fine.

My concern for 0.9.0 was overloading it with features, given Processors is quite the breaking-change.

mattbrailsford:
Well, if it's already a breaking change... :)

JimBobSquarePants:
@leekelleher I think we should ask around before merging this to see how many people are still relying on it while using v6.25+ I'm all about the latest and greatest but others are unable to upgrade, especially to not just a major version, but a few iterations on that v7.3.2? (That's a lot of regressive testing time).

If we make this breaking change we are essentially leaving them with a beta with no easy upgrade path. Somehow we made that mistake with my AzureProvider and now we've lost the ability to have new features tested which really sucks.

https://github.com/JimBobSquarePants/UmbracoFileSystemProviders.Azure/issues/26

mattbrailsford:
I really wish there was a way to know how many live instances are using Ditto and in what version. Maybe if HQ pinged back a list of DLL's as part of the update checker, you could find out, but I doubt that would happen.

The problem we have is that not that many vocal people seem to be using Ditto (I only know of 2 or 3) so it's hard getting the feedback. I think ultimately, not many people will actually bother to upgrade, as the processors thing is quite a big change (though is possible to upgrade).

I guess really it's a question of how many people still using 6.x over 7.x for new Umbraco sites?

JimBobSquarePants:
@mattbrailsford @leekelleher  We could always do this?

Branch off the latest v0.8.4 release. This branch is used exclusively as a base for bug fixes for people who cannot make the leap to Umbraco V7. As soon as v1.0 is released then we ditch that branch.

I'm up for adding Ditflow to v0.9.0 as long as that is the last major feature before v1.0 After that I'd really like to limit feature additions and concentrate very heavily on performance improvements and documentation.

mattbrailsford:
Sounds good to me. Can't see it changing drastically now, so happy to go
with that.

leekelleher:
:+1: For having Ditto 0.8.x for Umbraco v6.x support. I'm happy for bug fix / patch releases on that branch.

---

I'm sceptical to commit to saying this will be the last major feature before a v1.0.  While I don't envisage any other new features, it is impossible to say what we might come up with next.

e.g. I thought we had it nailed with our 0.8.0 release - I had no idea that Processors would be such a conceptual overhaul, but I'm totally glad we did it!

Re: Docs, stabilisation, performance :100:

mattbrailsford:
I know what you mean about changes, but I guess that's when we'd look
towards a V2. With the 0.8 branch we are already starting to maintain
backwards compatibility, so might as well move to a v1 and get people using
it.

leekelleher:
@mattbrailsford 1.0.0 before 0.9.0 release?

I'd prefer to get 0.9.0 released, let it settle/stabilise, then polish, etc. for a 1.0.0 release.

mattbrailsford:
Either / or, so long as there is a plan to move to 1.0.0. So if 0.9 is
Ditflo, and then we Polish / document for v1, that's all cool with me.

leekelleher:
OK, just agreed with @mattbrailsford about next versions roadmap.

* 0.9.0 - new Processors; compiles against Umbraco 6.2.5
* 0.10.0 - DitFlo feature; compiles against Umbraco 7.3.2

Everything that we said above about 0.8.x branch will apply to 0.9.x branch, we'll do bug fix patches for Umbraco 6.x support.

We'll look to get 0.9.0 release out as soon as possible :shipit:

leekelleher:
@mattbrailsford I've merged in the latest "develop" branch with this PR. I had to make a few amends, as mocking the `UmbracoContext` had changed somewhat (lets say evolved) in Umbraco v7.x.
We'll now use Moq for mocking the Umbraco services.

I'm happy for this to be merged in to the "develop" branch, for upcoming v0.10.0 release.

leekelleher:
Merged in for v0.10.0.

Issue #168, opened 2016-05-03
Mapping a polymorphic collection


leekelleher:
In issue #167, the topic of polymorphic collections came up, and would it be possible for DItto to map to them.

I've opened this as a separate thread so that we can discuss the concept, rather than implementation.

Do we think this is a good idea? Should we spend time to explore it?

As an example, I've put together an example of how it might work with Ditto.

> This takes inspiration from @micklaw's [Ditto Archetype Resolver feature for multiple-fieldsets](https://github.com/micklaw/Ditto.Resolvers#multiple-fieldsets).

```csharp
public interface IModel { }

[DittoDocType]
public class MyModel : IModel { }

[DittoDocType("myDocType2")]
public class MyModel2 : IModel { }

[Test]
public void Polymorphic_Collection_Maps()
{
    var nodes = new List<PublishedContentMock>()
    {
        new PublishedContentMock { Id = 1111, DocumentTypeAlias = "myDocType1" },
        new PublishedContentMock { Id = 2222, DocumentTypeAlias = "myDocType2" },
        new PublishedContentMock { Id = 3333, DocumentTypeAlias = "myDocType3" },
    };

    var items = nodes.As<IEnumerable<IModel>>();

    Assert.That(items, Is.Not.Null);
    Assert.That(items.Count(), Is.EqualTo(3));

    CollectionAssert.AllItemsAreInstancesOfType(items, typeof(IModel));

    Assert.That(items.First(), Is.TypeOf<MyModel>());
    Assert.That(items.Skip(1).First(), Is.TypeOf<MyModel2>());
    Assert.That(items.Skip(2).First(), Is.TypeOf<MyModel>());
}
```

I've posted this code as a gist too: <https://gist.github.com/leekelleher/0b1a9b9a0623e997031b8ac65f19d1c5>

Nicholas-Westby:
> Do we think this is a good idea? Should we spend time to explore it?

Yep and yep :smile:

Here is one use case, assuming you are mapping content nodes:

> I have a collection of `IArticle`. I want to use Ditto to populate that collection using instances of the classes `Article`, `BlogPost`, `MarkdownPost`, `RichTextPost`, `PressRelease`, and `NewsPost`. The Umbraco document type alias can be used to disambiguate between the items to know which class to instantiate.

Here is another use case, assuming you are mapping Archetype fieldsets:

[...]

PR #169, opened 2016-05-04
DittoFactoryAttribute

mattbrailsford:
Created a DittoFactory processor attribute to allow the dynamic typing of items based upon properties of that item, such as Nested Content items which support a base interface, but may have differing concentrate models named after their DocType alias.

I think more unit tests are needed, and I'm not 100% certain in my head if this is entirely needed, but I put together a POC anyway.

This would be a fix for issue #168 if it was merged in.


src/Our.Umbraco.Ditto/ComponentModel/Processors/DittoFactoryAttribute.cs
```diff
@@ -0,0 +1,82 @@
...
+            var propType = Context.PropertyDescriptor.PropertyType;
+            var propTypeIsEnumerable = propType.IsEnumerableType();
+            var baseType = propTypeIsEnumerable ? propType.GetEnumerableType() : propType;
+
+            // TODO: Validate the base type more?
+
+            // Find the appropreate types
+            // There is no non generic version of ResolveTypes so we have to
+            // call it via reflection
+            var method = typeof(PluginManager).GetMethod("ResolveTypes");
```
JimBobSquarePants: We should save this as a compiled delegate to save looking it up every time.

src/Our.Umbraco.Ditto/ComponentModel/Processors/DittoFactoryAttribute.cs
```diff
@@ -0,0 +1,82 @@
...
+            if (enumerableValue != null)
+            {
+                var items = enumerableValue.Select(x =>
+                {
+                    var typeName = typeNameResolver.ResolveTypeName(Context, x);
+                    var type = types.FirstOrDefault(y => y.Name.InvariantEquals(typeName));
+
+                    // QUESTION: Should we return null? throw an exception? or strip items if a model can't be found?
+
+                    return type != null ? x.As(type): null;
```
JimBobSquarePants: I vote null rather than strip.

JimBobSquarePants:
Will need to download and test the code but it looks good to me.

src/Our.Umbraco.Ditto/ComponentModel/Processors/DittoFactoryAttribute.cs
mattbrailsford: Feel free to weak as I'm not entirely sure how you do those.

JimBobSquarePants: magic :wink:

mattbrailsford: I have no doubt :8ball:

mattbrailsford:
Wondering if I should drop the use of DittoFactoryTypeNameResolver and instead just make DittoFactoryAttribute abstract and have an abstract method on this called ResolveTypeName(IPublishedContent currentContent)?

It does mean that you couldn't use DittoFactory directly as an attribute, passing a type, but it would mean you can pass additional config in, for example, with the DittoDocTypeFactory, you might want to accept a "Prefix" / "Suffix" parameters so you can easily append something extra on the doctype alias? And I don't see this being any less / more work as you would have had to have defined a custom DittoFactoryTypeNameResolver anywho.

Thoughts?

Nicholas-Westby:
Would love to test this out in a starter kit I'm building for my company. What would be the easiest way of going about that?

Should I clone the repo, switch to the `feature/ditto-factory` branch, and then build the code somehow (I'm guessing you have some streamlined way to do this)? Does building generate binaries or a NuGet package?

mattbrailsford:
You can grab the latest build from the build server
https://ci.appveyor.com/project/leekelleher/umbraco-ditto/build/0.10.0.521/artifacts

mattbrailsford:
Have dropped the DittoFactoryTypeNameResolvers and just gone with an abstract DittoFactoryAttribute instead. Less moving parts, and more options.

leekelleher:
@mattbrailsford All cool with the class name change.

leekelleher:
I've spent some time today going over the codebase, (fixing up StyleCop stuff :smile_cat:; refactoring some unit-tests, etc).  I'm totally happy with this PR being merged in and we can make further unit-tests from the "develop" branch.

@JimBobSquarePants @mattbrailsford Does that sound okay?

mattbrailsford:
Cool with me

JimBobSquarePants:
YASSSS STYLEKITTEN IS A THING NOW! :smile_cat: 

Ahem... Yeah good to go.

Issue #157, opened 2016-02-25
API reference documentation


leekelleher:
Since we have such lovely XML comments through our code-base, should we look to auto-generate static API reference? Do we think this would be useful?

For my client work-projects, I've tried Sandcastle ([SHFB](https://github.com/EWSoftware/SHFB)) and [Doxygen](http://www.stack.nl/~dimitri/doxygen/), both produce good results.

Yesterday I was told about [DocFX](https://github.com/dotnet/docfx), which is being used for new .NET Core docs, (and is supported by the .NET Foundation).  I'd be interested in trying it out on Ditto.

We could wire it up with AppVeyor builds - pushing up the latest API ref docs to the GitHub release pages.

JimBobSquarePants:
I made a poor man attempt at using DocFX with ImageProcesorCore but couldn't get it working. (I have an open issue for it https://github.com/JimBobSquarePants/ImageProcessor/issues/297). 

I'm dying to give it a go. The Core docs are sweet! http://dotnet.github.io/api/index.html

leekelleher:
On a side-note, I've been using (and contributing to) [Wyam](http://wyam.io/) recently. It has a [static code analysis module](http://wyam.io/modules/analyzecsharp) that can be used to generate API reference docs.

leekelleher:
Quick update to say that I've started prototyping our docs using Wyam. I haven't pushed my local branch up yet, I'll see how I progress over the next few days.

My biggest concern is the front-end dev, CSS, etc. I'm tempted to do it in vanilla HTML and let you guys work your magic! :wink:

mattbrailsford:
Something like this do?

![capture](https://cloud.githubusercontent.com/assets/527305/15077208/4ad79fe8-13a5-11e6-8ab9-80f2d9aa2ee3.PNG)

leekelleher:
@mattbrailsford Nice!!

mattbrailsford:
Had a look at the docs with a view to coming up with a better structure. What do we think of the following:

- Installation
  - Requirements
  - Our Umbraco Instalation
  - Nuget Instalation
  - Manual Installing
  - Upgrading to 0.9?
- Usage Guide
  - Basic Usage
  - Processors
    - In-built processors
  - Custom Processors
    - Processor Contexts (Advanced?)
  - Chaining Processors
  - Caching Processors
    - Cache Key Builders
  - Conversion Handlers
  - DittoView
  - DittoController (Advanced?)
[...]

PR #170, opened 2016-05-06
API Documentation using Wyam

leekelleher:
As mentioned on #157, I've started prototyping our API documentation using Wyam.

This is completely open to discussion, as I want us all to be comfortable with it.

If you want to try it out locally, then here's what to do...

1. Checkout the "docs/wyam-prototype" branch
2. Go to the "/docs" folder...
  a. Run the "get-wyam.cmd" batch file, (this will download Wyam via NuGet)
  b. From a command-prompt, (in the "/docs" folder), run this...
    * `.wyam\Wyam.exe --watch --preview`

  c. Open a browser at <http://localhost:5080/>

I haven't added any CSS styling. For the markup, I've mostly took inspiration from the [Wyam.io](http://wyam.io) docs.

Like I say, this is totally open for discussion, I'm a believer in Cunningham's Law! :stuck_out_tongue_closed_eyes:


leekelleher:
Given that GitHub Pages can now generate from the `/docs` folder, ([see their blog post about it](https://github.com/blog/2233-publish-your-project-documentation-with-github-pages)). We could switch back to using GitHub Pages/Jekyll for handling the docs?

As much as I like Wyam, I appreciate that it is another layer to the build process.

Thoughts?

mattbrailsford:
After playing with wyam, I'd be inclined to agree. I think the only benefit it brings is being able to generate the API docs, but I'm guessing we might still be able to try and automate that in some way. Even if not, I think simplicity has to be key otherwise we just won't keep up to it.

mattbrailsford:
If we do close this PR though, it's whether we need to re-open an issue to write the docs :)

leekelleher:
Yeah, we'll close this one off, (no immediate rush though), but let's keep the branch until we've made a decision and/or have done the switch over. I'll create a separate issue for the Jekyll-based docs (or re-open the old one)

leekelleher:
Closing this PR. Raised new issue #192

PR #180, opened 2016-07-17
Fix #179

### LINKED ISSUE #179: MultinodeTreePicker behaviour changed in 0.9.x
In previous versions assigning a `DittoPickerConverter` to a `MyClass` would always return an `IEnumberable<MyClass>` whether one or many items were picked.

In 0.9.0 when a single item is picked the following exception is thrown.

> Additional information: Unable to cast object of type 'System.String[]' to type 'System.Collections.Generic.IEnumerable`1[MyClass]'.

Umbraco Version 7.4.3

JimBobSquarePants:
Adds the ability to track attributes on typeparams in generic types.
This matches behaviour in pre 0.9.0 releases.


mattbrailsford:
With this, do we assume that the processor is capable of handling both singular and enumerable input values?

Issue #184, opened 2016-07-28
Allow UmbracoProperty attribute to choose between own property vs properties collection mapping


mattbrailsford:
Currently when mapping a property we first check own properties defined on the IPublishedContent instance and if a suitable property isn't found, we check the properties collection instead. The issue is for things like "name", "children", "version" etc, all of which are defined by default on the IPublishedContent, these will hide any properties found in the properties collection with those names so we need to find a way to allow these properties to be mapped in some way.

mattbrailsford:
There is some additional background information in issue #181

JimBobSquarePants:
I'm struggling to get my head around this one. Why would someone use a property name in the collection that matches a property name on `IPublishedContent`? To me that's asking for trouble.

My thought would be to check if the type implements IPublished content and if so then skip the first check.

mattbrailsford:
@JimBobSquarePants that was my thought as well really. I do have a local branch where I have implemented some code that, if in debug mode, and when mapping a declared property, if that property is part of IPublishedContent (ie, a property outside the control of the developer) and there is a property existing in the properties collection with that alias that it will write a log message to tell you the fact.

Personally, I'm thinking this is good enough as like you, I think it's a bad idea to be naming properties with the same alias of those reserved property names, but it at least gives the dev something to go off if they do hit a problem.

mattbrailsford:
I've created a branch an commited some code here fc981f8 which performs that check on property being mapped to see if it is an IPublishedContent property and b) if it is hiding an umbraco property and if so logs an warning. There is quite a lot of code in there, but much of it is just to allow me to unit test the fact it works so I need to fake debug mode and also mock the logger so I can confirm the message gets logged.

jamiepollock:
@mattbrailsford @JimBobSquarePants I have to agree with you both here. I guess this is also to allow for properties Model Builder classes? Those get passed in as IPublishedContent inherited classes to Ditto when that's enabled.

Nicholas-Westby:
@JimBobSquarePants To answer your question, this happened to both me and a colleague of mine, so it can't be that rare of a situation. Here are some of the properties on IPublishedContent I could see tripping people up:

* Children
* CreateDate
* CreatorName
* Id
* ItemType
* Level
* Name
* Parent
* Path
* Properties
* SortOrder
* UpdateDate
* Url
* Version
* WriterName
* Item (from the indexer)

[...]

PR #185, opened 2016-07-29
Adds PropertySource to UmbracoPropertyAttribute 

mattbrailsford:
This is a fix for issue #184. 

Added code to allow you to select a property source when using UmbracoPropertyAttribute so that you can define the order in which properties are checked between the instance properties and properties in the umbraco properties collection. This can be set per property using the PropertySource parameter of the UmbracoPropertyAttribute, or globally by setting Ditto.DefaultPropertySource in an app startup handler. If using either InstanceThenUmbracoProperties (the default) or UmbracoThenInstanceProperties and the site is in debug mode, and ditto detects that the property requested hides a property in the other source, then a warning log message will be logged to the log file giving instructions to either not use reserved aliases, or to use PropertySource to override the behavior if required.

### DIFF
--- a/src/Our.Umbraco.Ditto/Common/DittoDisposableTimer.cs
+++ b/src/Our.Umbraco.Ditto/Common/DittoDisposableTimer.cs
@@ -22,9 +22,9 @@ namespace Our.Umbraco.Ditto
         /// <typeparam name="T">The type of the calling class.</typeparam>
         /// <param name="startMessage">The starting message for the profiler.</param>
         /// <returns>Returns an instance of the disposable timer.</returns>
-        public static new DisposableTimer DebugDuration<T>(string startMessage)
+        public new static DisposableTimer DebugDuration<T>(string startMessage)
         {
-            if (Ditto.IsDebuggingEnabled)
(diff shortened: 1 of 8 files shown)


Nicholas-Westby:
Seems like a very reasonable solution. I like it #h5yr

JimBobSquarePants:
Looks great to me. Love that you've standardized the property flags and handled the indexer also. Very robust!

jamiepollock:
Agreed, this is fantastic. I was thinking one feature I might expect would be to set at class level using `UmbracoProperties` attribute the PropertySource for a specific class. This would be different from at property level or the global default.

What does everyone thing? Separate ticket or a no goer?

mattbrailsford:
@jamiepollock not a bad idea, and pretty easy to implement. Give me a minute :)

mattbrailsford:
Ok, maybe 6 minutes :)

jamiepollock:
Great work is still great work, chap. Bingo bango! I feel like this pretty much complete now :) #h5yr!

mattbrailsford:
Just gonna do a unit test around the UmbracoProperties stuff, then yea, I'm happy to merge in. Thanks for all the feedback everyone.

Issue #177, opened 2016-06-14
Problem with overridden properties that depend on other properties


Slowacki:
I have a following-ish setup in my project. 

An abstract class with some virtual properties:
```
[JsonObject(MemberSerialization.OptIn)]
    public abstract class Base: IInterface1, IInterface2
    {
        [JsonProperty(PropertyName = "id")]
        public virtual string IdString { get { return IdInt.ToString(); } }

        [JsonProperty]
        public virtual int IdInt { get; set; }
    }
```

And a class that inherits it:
   ```
 [JsonObject(MemberSerialization.OptIn)]
    public class NewClass : Base
    {    
        [JsonProperty(PropertyName = "id2")]
        public override string IdString { get { return IdInt.ToString(); } }

        [JsonProperty("id1")]
        [TypeConverter(typeof(IdContentPickerConverter))]
        [UmbracoProperty("ItemLink")]
        public override int IdInt { get; set; }
    }
```

After mapping IPublishedContent to NewClass, what I would expect is to have IdInt and IdString output as the same value, but one as an int and the other as a string. However, unfortunately that is not what happens. The output looks like that:
```
{
id2: "0",
id1: 1613
}
```

If I instead create a new property instead of IdString in the NewClass or change the modifier from `override` to `new` (which I obviously don't want to do), the IdString value is output correctly.

I'm using Ditto v0.8.3 and Umbraco v7.3.7.

leekelleher:
@Slowacki Thanks for raising this.  Ditto shouldn't be attempting to do anything with the `IdString` property, since it doesn't have a setter method.  But I'll set up a unit-test to try out this scenario.

leekelleher:
@Slowacki OK, I can reproduce the error, it seems to be related to using the `virtual` properties within the abstract `Base` class.

We'll probably need @JimBobSquarePants to take a look at this too. Here's a gist of the unit-test I made (branched off from "0.8.4" tag)...

<https://gist.github.com/leekelleher/5985a13e6716d77c2b5d0790b7db21d2>

JimBobSquarePants:
@leekelleher What does v0.9.0 do in this scenario?

leekelleher:
@JimBobSquarePants I haven't tried it yet... will do (between CG16 sessions) :metal:

leekelleher:
@JimBobSquarePants Tried it... it's the same issue on v0.9.0.
(_No rush on this though, we're at CG16 after all_ :smirk_cat:)

[...]

PR #188, opened 2016-09-04
Proxy class properties accessing object instance properties

leekelleher:
Added failing unit-test for issue #177

When using proxy classes, it appears that lazy-loaded properties can not access the property values of its object instance.

@JimBobSquarePants to review (in due course - no rush, just wanted to keep track of this) :sunglasses:

### DIFF
--- a/tests/Our.Umbraco.Ditto.Tests/Issue177Tests.cs
+++ b/tests/Our.Umbraco.Ditto.Tests/Issue177Tests.cs
@@ -1,0 +1,81 @@
+using NUnit.Framework;
+using Our.Umbraco.Ditto.Tests.Mocks;
+
+namespace Our.Umbraco.Ditto.Tests
+{
+    /// <remarks>
+    /// Unit-test to verify an issue - as outlined in #177
+    /// https://github.com/leekelleher/umbraco-ditto/issues/177
+    /// When a property is marked as `virtual`, it becomes lazy-loaded,
+    /// and not accessible by other properties on the model.
+    /// </remarks>
+    [TestFixture]
+    public class Issue177Tests
+    {
+        interface IInterface1
+        {
+            int IdInt { get; set; }
+        }
+
+        interface IInterface2
+        {
+            string IdString { get; }
+        }
+
+        public abstract class Base : IInterface1, IInterface2
+        {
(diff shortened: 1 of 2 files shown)


mattbrailsford:
Just to note what lee and I observed when we reviewed this (to the best of our understanding), it looks like an IProxy object is made per property that only contains that individual property, so if one property accesses another, it doesn't exist on the IProxy, so it fails.

JimBobSquarePants:
@mattbrailsford No, that's not correct at all. An `IProxy` is created per instance.

https://github.com/leekelleher/umbraco-ditto/blob/develop/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs#L362

Basically if you have a class like this:

``` c#
public class Foo {
    public virtual string Bar {get; set;}
}
```

The proxy generator will create a class like this:

``` c#
public class FooProxy : Foo, IProxy {

    public IInterceptor Interceptor { get; set; }
    
    public override string Bar {
        get {
            return (string) interceptor.Intercept(methodof(Foo.get_Property), null);
        }
        set {
            interceptor.Intercept(methodof(Foo.set_Property), value);
        }
}
```

We set the interceptor containing all the virtual properties [here](https://github.com/leekelleher/umbraco-ditto/blob/develop/src/Our.Umbraco.Ditto/Proxy/ProxyFactory.cs#L44).  It also reference the current instance. 

In the example code the lazy property seems to be getting called against the base class and not the proxy class but Ill have to dig around to be sure. Now we have a failing test I should easily be able to do so. 

It really is an odd design pattern though. Why have to properties referencing the same object? If they want to convert it to a string they should be using a method or casting their property. They're hurting performance and making their code difficult to follow.

mattbrailsford:
Is this still relevant with PR #191?

JimBobSquarePants:
TBH I don't know. Will have to conduct further tests after we merge.

leekelleher:
Merged in and tested locally... the failing unit-test (for issue #177) now passes.

Merged in this PR - we can decide whether or not to keep the actual unit-test, or to combine with another unit-test (from PR #190)?

JimBobSquarePants:
I'd move the tests together. still want to have a dig around though.

Issue #190, opened 2016-09-05
Changes to lazy properties


mattbrailsford:
In the current version of ditto, a lazy property is defined purely by being virtual. A problem with this is that if you are using a base class with overridable properties, this property needs to be virtual and in so doing, becomes lazy.

Now, there are 2 things I want to discuss here:

1) We keep lazy properties but maybe make them more explicit, maybe by using a new attribute such as

    [DittoLazy]
    public virtual MyPRop { get; set; }

This way, it's clear that you intend this property to be lazy, rather than just assuming all virtual properties are lazy.

2) The more controversial part of the question is, do we need lazy properties at all? 

This feature was introduced early on in ditto's life, when it's place in the world was still being discovered. Now though, we know that we want ditto to be a view models builder, and not just a general mapper. Given that the idea behind a view model is to only make available to the view the properties that are needed by the view, the question is why should lazy properties be needed at all? Because, if we have declared a property on our view model, then it should always be needed and thus there is no need to be lazy.

mattbrailsford:
This to some degree relates to pr #188

JimBobSquarePants:
> 2) The more controversial part of the question is, do we need lazy properties at all?

We need them. I use them constantly and would recommend most people do. Not everyone uses Ditto with view models. I don't and I'd much prefer not to. I have a well established process which works perfectly and allows me to very rapidly create performant websites.

I really don't like the idea of the `[DittoLazy]` attribute at all. The `virtual` keyword is clear enough and is used by every proxy generator out there (Castle, Entity Framework) for interception. Adding an attribute simply adds overhead. 

Devs shouldn't be using virtual properties in ViewModels anyway since Viewodels shouldn't be inheriting from each other. It's pollutant and screams of poor design. If any inheritance is to take place it should be from a single base class which shouldn't require overwriting if designed appropriately.

mattbrailsford:
Hmm, can't help but feel that sounds a little double standards to say we should support lazy when arguably you shouldn't have properties on your view model you don't need, but then to say you shouldn't be using virtual, overridable properties on your view model. Surely we should be either flexible, or opinionated, but to mix the two sounds a little too contradictory. 

[...]

PR #191, opened 2016-09-16
DittoLazy attribute

mattbrailsford:
As discussed in #190, adding a DittoLazy attribute to be explicit about which properties are Lazy loaded.

@JimBobSquarePants can you check at line 385 that this is ok? In the original code, virtual props were made first, and this code ran, THEN non virtual property values were set. Should this be ok that all properties are set first lazy and non lazy, and THEN the proxy object is created?

Still need to do some unit tests

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoLazyAttribute.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoLazyAttribute.cs
@@ -1,0 +1,12 @@
+﻿using System;
+
+namespace Our.Umbraco.Ditto.ComponentModel.Attributes
+{
+    /// <summary>
+    /// The Ditto lazy property attribute.
+    /// Used for specifying that Ditto should lazy load this property.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+    public class DittoLazyAttribute : Attribute
+    { }
+}
--- a/src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoLazyAttribute.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoLazyAttribute.cs
@@ -6,7 +6,7 @@ namespace Our.Umbraco.Ditto.ComponentModel.Attributes
     /// The Ditto lazy property attribute.
     /// Used for specifying that Ditto should lazy load this property.
     /// </summary>
-    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+    [AttributeUsage(AttributeTargets.Property | AttributeTargets.Class, AllowMultiple = false)]
     public class DittoLazyAttribute : Attribute
     { }
 }
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -332,54 +332,31 @@ namespace Our.Umbraco.Ditto
             // Process all the properties.
             if (properties.Any())
             {
-                foreach (var propertyInfo in properties)
+                // process lazy load properties first
+                foreach (var propertyInfo in properties.Where(x => x.ShouldAttemptLazyLoad()))
                 {
-                    var propLazyAttr = propertyInfo.GetCustomAttribute<DittoLazyAttribute>();
-                    if (propLazyAttr != null || ((typeLazyAttr != null || Ditto.LazyLoadStrategy == LazyLoad.AllVirtuals) && propertyInfo.IsVirtualAndOverridable()))
-                    { 
-                        // Configure lazy properties
-                        using (DittoDisposableTimer.DebugDuration<object>(string.Format("ForEach Lazy Property ({1} {0})", propertyInfo.Name, content.Id)))
-                        {
-                            // Ensure it's a virtual property (Only relevant to property level lazy loads)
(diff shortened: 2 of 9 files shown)


src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoLazyAttribute.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Our.Umbraco.Ditto.ComponentModel.Attributes
+{
+    /// <summary>
+    /// The Ditto lazy property attribute.
+    /// Used for specifying that Ditto should lazy load this property.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property | AttributeTargets.Class, AllowMultiple = false)]
```
JimBobSquarePants: Allow on enum and interface too. We have a constant now don't we for this?

There's also a way we can globally register isn't there?

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -310,84 +305,93 @@ public static class PublishedContentExtensions
...
-                        if (ignoreAttr != null)
+                    var propLazyAttr = propertyInfo.GetCustomAttribute<DittoLazyAttribute>();
+                    if (propLazyAttr != null || (typeLazyAttr != null && propertyInfo.IsVirtualAndOverridable()))
+                    { 
+                        // Configure lazy properties
+                        using (DittoDisposableTimer.DebugDuration<object>(string.Format("ForEach Lazy Property ({1} {0})", propertyInfo.Name, content.Id)))
                         {
-                            continue;
+                            // Ensure it's a virtual property (Only relevant to property level lazy loads)
+                            if (!propertyInfo.IsVirtualAndOverridable())
```
JimBobSquarePants: Good check.

src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoLazyAttribute.cs
mattbrailsford: I think there is a confusion I need to check here. Having the attribute declared on a type currently makes all of it's contained virtual properties lazy, so it could make sense to use on an Interface (assuming you can declare properties as virtual on an interface) but wouldn't so much on an Enum.

I did have this discussion with Lee, but would be good to get your view point. What are you expecting the class level Lazy attribute to do? Make all it's inner properties lazy? or make the parent declaring property lazy? (ie, the parent property that defines it's type as your type with the lazy attribute applied). If it's the later, how would this affect enumerable props?

mattbrailsford:
Thanks @JimBobSquarePants. We don't have the global register / global flag yet, but should be easy enough to add.

src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoLazyAttribute.cs
JimBobSquarePants: I'd expect it to work in the same way as other attributes. I.E if I mark an enum with the `EnumAttribute`  then every time I declare that enum type as a property I know the processor will kick in without me having to declare the attribute again. Same as with the `UmbracoPickerAttribute`. I declare any base classes with that attribute so I can use them as properties at any time and always have a working picker.

JimBobSquarePants: Id expect it to work in the same way as other attributes. I.E if I mark an enum with the enum processor then every time I declare that enum type as a property I know the processor will kick in without me having to declare the attribute again...

.... Damn your brain! I really like that you can declare all properties within a class as lazy. thinking about it that's a really neat way to do it.

JimBobSquarePants: Damn your brain! Now I think about it really like that you can declare all properties within a class as lazy. That's a really neat way to do it and way better than what I was thinking.

mattbrailsford: Ok, cool, well worth f you are happy with way too, let's go with it. I think it's the easiest to comprehend.

mattbrailsford: Oops, that should have been "well if you are happy"

mattbrailsford: With this approach, would we still want a Ditto.RegisterLazyAttribute function to register the class level ditto lazy on types you may not have control over? (again, this would just make that types inner virtual properties lazy, not the type itself)

mattbrailsford:
Added some starter unit tests but I could do with some advice from @JimBobSquarePants as to how we can verify in code that a given property has an interceptor registered?

src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoLazyAttribute.cs
JimBobSquarePants: I don't think we'll need to. There's plenty of fine grained control already now.

JimBobSquarePants:
@mattbrailsford You should be able to check whether the result implements `IProxy`

mattbrailsford:
Yea, currently doing that, but no easy way to "check" which properties are
setup to be proxied.

Issue #176, opened 2016-06-12
InvalidCastException


sniffdk:
Hi Ditto-people

I've just upgraded to Ditto 0.9.0 and have started seeing an issue where an InvalidCastException is thrown, something like this:
`Unable to cast object of type 'Web.ViewModels.ItemViewModel[]' to type 'System.Collections.Generic.List`1[Web.ViewModels.ItemViewModel]'.`

I'm using Ditto in conjunction with Nested Content doing some fairly simple stuff like:

```
var items = publishedNode
    .GetPropertyValue<IEnumerable<IPublishedContent>>("items")
    .As<ItemViewModel>()
    .ToList();
```

I tried to remove the `.ToList()` part and run it through a manual for-each, but the exception is still thrown.
The Ditto stack trace looks like this:

```
at lambda_method(Closure , Object , Object )
at Our.Umbraco.Ditto.PropertyInfoInvocations.SetValue(PropertyInfo property, Object instance, Object value)
at Our.Umbraco.Ditto.PublishedContentExtensions.ConvertContent(IPublishedContent content, Type type, CultureInfo culture, Object instance, IEnumerable`1 processorContexts, Action`1 onConverting, Action`1 onConverted)
at Our.Umbraco.Ditto.PublishedContentExtensions.As(IPublishedContent content, Type type, CultureInfo culture, Object instance, IEnumerable`1 processorContexts, Action`1 onConverting, Action`1 onConverted)
at Our.Umbraco.Ditto.PublishedContentExtensions.<>c__DisplayClass3.<As>b__1(IPublishedContent x)
```

And apparently the Source is mentioned as being `Anonymously Hosted DynamicMethods Assembly`, gotta :heart: dynamics :smile: 
This is tested in Umbraco 7.4.3, both on .net 4.5.2 and 4.6.1 and downgrading to Ditto 0.8.4 made the exception go away.

I don't know anything about the Ditto internals, but do let me know if I can provide any more help or clues.

Cheers
Mads

sniffdk:
Ok, did some more testing. First a small sample of the model setup I'm using:

```
public class ListViewModel
{    
    public string Name { get; set; }
    ...........

    public List<ItemViewModel> Items { get; set; }

    public ListViewModel()
    {
        Items = new List<ItemViewModel>();
    }
}
```

Now, this fails in version 0.9.0 on `PropertyInfoInvocations.SetValue` which is some of the new performance optimizing stuff.
[...]

PR #195, opened 2016-10-11
EnumerableConverter fails with inherited class

leekelleher:
An issue has been flagged for when attempting to map a property type that inherits from a generic enumerable.

Background on the issue can be found here:
https://our.umbraco.org/projects/developer-tools/ditto/ditto-feedback/79244-error-on-using-relatedlink-with-ditto-and-jeavon-property-converter#comment-257744

The culprit is here: [EnumerableConverterAttribute.cs, line 76](https://github.com/leekelleher/umbraco-ditto/blob/8387c47bc5f0a2c01ed5b3299506f91c28bd4234/src/Our.Umbraco.Ditto/ComponentModel/Processors/EnumerableConverterAttribute.cs#L76)
As the object-type is detected as an enumerable type ([line 40](https://github.com/leekelleher/umbraco-ditto/blob/8387c47bc5f0a2c01ed5b3299506f91c28bd4234/src/Our.Umbraco.Ditto/ComponentModel/Processors/EnumerableConverterAttribute.cs#L40)), but when the `GenericTypeArguments` are accessed ([line 76](https://github.com/leekelleher/umbraco-ditto/blob/8387c47bc5f0a2c01ed5b3299506f91c28bd4234/src/Our.Umbraco.Ditto/ComponentModel/Processors/EnumerableConverterAttribute.cs#L76)),
the top-level type is accessed, not the base type.

These unit-tests attempt to test the processor directly, (as opposed to going via the `.As<T>` method.
The first test is successful, as it is working with a generic enumerable type directly.
The second test fails, as it is working with a class inherited from a generic enumerable type.


JimBobSquarePants:
@leekelleher  Interesting... Reading through the issue it does seem like the PropertyValueConverter library should not be returning null. That would fix the issue yeah?

If so I don't really want to mess around with our code as returning null instead of an empty enumerable is considered poor practise and I don't want to be hacking around to deal with that.

leekelleher:
@JimBobSquarePants I agree that the PropertyValueConverter shouldn't return a null. 👍 

As for the isolated unit-test itself, would we expect that to work or not?

If an object inherits from a generic enumerable type, should [line 40](https://github.com/leekelleher/umbraco-ditto/blob/8387c47bc5f0a2c01ed5b3299506f91c28bd4234/src/Our.Umbraco.Ditto/ComponentModel/Processors/EnumerableConverterAttribute.cs#L40) detect that it is an `IEnumerable`? and then should [line 76](https://github.com/leekelleher/umbraco-ditto/blob/8387c47bc5f0a2c01ed5b3299506f91c28bd4234/src/Our.Umbraco.Ditto/ComponentModel/Processors/EnumerableConverterAttribute.cs#L76) be able to find the first `GenericTypeArguments` type?

If the answer is yes, then I guess there's another question about how deep down the inheritance should we check?

JimBobSquarePants:
Ok @leekelleher... So I've finally had the chance to put some thought to this.

The returned value should be null but we should be able to correctly detect and return that value without throwing. 

I've updated the PR with my updated fix. 

Basically the logic is as follows.
1. If the return type is `IEnumerable<T>` we can cast and return.
2. If the return type implements `IEnumerable<T>` and has an empty constructor we can create an instance and return.
3. If the return type implements `IEnumerable<T>` and does not have an empty constructor we return null.

BTW Having done this digging I finally fixed #176 also.

We now need to do a PR in Umbraco-Core-Property-Value-Converters to add an empty constructor and return an new empty instance instead of null.

JimBobSquarePants:
@leekelleher  I added a test that initially failed in the PR 😄 

https://github.com/leekelleher/umbraco-ditto/pull/195/files#diff-dcacc48ee99c56708eba88c3efecd08bR90

leekelleher:
@JimBobSquarePants Bad news, it turns out that our AppVeyor hasn't been running our unit-tests for the past 11 months, (because we renamed the assembly and didn't update the appveyor.yml), _anyhooo_ ... it turns out that there's a few failing unit-tests...

- EnumerableConverter_EnumerableModelEmptyDictionary_Test
- EnumerablesCast
- GenericDictionaryPropertyIsNotDetectedAsCastableEnumerable

I was going to start taking a look at them, but it's getting pretty late tonight (UK time) ... and you'll probably a better idea of how to tackle the `Dictionary<K, V>` stuff 😼

JimBobSquarePants:
Really? I ran them all locally and everything was passing. Will have a look.

JimBobSquarePants:
**EnumerableDetectionTests.GenericDictionaryPropertyIsNotDetectedAsCastableEnumerable** 

Directly contradicts with 

**TypeInferenceTests.TestIsCastableEnumerableType**

We're expecting **not null** with the first but **false** in the second.

The source for the latter deliberately excludes dictionary since we'd have to go further down the rabbit hole to discover what the subtypes are. My mantra was, if you are expecting a dictionary then you are using a `PropertyValueConverter `which would return a dictionary type.

```
public static bool IsCastableEnumerableType(this Type type)
{
    // String, though enumerable have no generic arguments.
    // Types with more than one generic argument cannot be cast. 
    // Dictionary, though enumerable, requires linq to convert and shouldn't be attempted anyway.
    return type.IsEnumerableType() && type.GenericTypeArguments.Any()
            && type.GenericTypeArguments.Length == 1
            && type.TryGetElementType(typeof(IDictionary<,>)) == null;
}
```
I originally wrote it nearly 450 days ago so something has definitely got lost in translation when we migrated the types as it's never been a feature.

Sooooo.... Do we try to figure it out?

leekelleher:
@JimBobSquarePants Are you happy for me to merge this in? or did you need help figuring out why the `Factory_Resolves` test fails? (we could open as a separate issue)

💯 I agree about commenting our unit-tests... same old story, at the time of writing the test I think it makes sense, then later forget what it does. I also think we have a lot of duplication in our tests - but I'm hesitant to refactor/remove any at the moment.

JimBobSquarePants:
Yeah, very happy to merge. The other issue was thrown by umbraco not us so I'll but that aside for now.

Let's definitely not remove tests just do a little tidy up with comments/formatting.

JimBobSquarePants:
Yay!

PR #197, opened 2016-10-21
Added some more caching options, and cache key customisation

tompipe:
Added some new caching options.

CacheBy.AttributeType - Uses the type.FullName to cache the attribute value
CacheBy.Custom - Uses a method on the attribute to allow further customisation of the cache key

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoCacheableAttribute.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoCacheableAttribute.cs
@@ -58,18 +58,27 @@ namespace Our.Umbraco.Ditto
                 return refresher();
             }
 
-            // Get the cache key builder type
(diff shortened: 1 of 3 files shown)


mattbrailsford:
Not sure why we'd use custom over a cache key builder? Is it simply to not have to define another class?

leekelleher:
@tompipe Follow up on this - just bumping for @mattbrailsford's question really.

tompipe:
Oops - missed that. Apologies. Let me dig out the code and refresh my memory.

leekelleher:
Bumping - @tompipe Is this PR still relevant? Thanks.

tompipe:
Hmm, I still haven't had time to check! 😕

Im gonna say probably not, but I'll look back when I get home

tompipe:
Yikes - I've been useless with looking at this haven't I. Sorry :(

I'll reopen it, if i ever get time to dig out the code! Cheers

PR #199, opened 2016-12-01
RecursiveDitto processor doesn't maintain the ProcessorContexts

leekelleher:
If a processor returns an `IPublishedContent`, then the recursive processor (one of the default post-processors) will make an `.As()` call, but it doesn't pass in current `DittoProcessorContext` objects.  This pull-request aims to resolve this.

@mattbrailsford Are you happy with the processor contexts being passed through to `GetPostProcessorAttributes()`?


mattbrailsford:
Yup

Issue #200, opened 2017-01-06
Base Processor - exposing Umbraco context services


leekelleher:
From reading Umbraco core's recent [Common Pitfalls & Anti-Patterns](https://our.umbraco.org/documentation/Reference/Common-Pitfalls/) document, should we consider exposing the Umbraco objects via our [`DittoProcessorAttribute`](https://github.com/leekelleher/umbraco-ditto/blob/develop/src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoProcessorAttribute.cs)? (rather than letting developers use the static singletons)

Of course, under the hood we'd actually be exposing the static singletons, (_since this is how Umbraco core do it - e.g. [`SurfaceControllerFactory.cs`](https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Web/Mvc/SurfaceControllerFactory.cs#L56); not using dependency injection_), this would encourage developers to use our exposed properties, so in future we could potentially replace/inject them (or however Umbraco core team recommend we package devs should do this).

Based on the doc, (and the [Umbraco core's `PluginController` code](https://github.com/umbraco/Umbraco-CMS/blob/dev-v7/src/Umbraco.Web/Mvc/PluginController.cs)), we could expose the following properties...

- `ApplicationContext`
- `UmbracoContext` exposed as `Umbraco`
- `DatabaseContext`
- `ServiceContext` exposed as `Services`
- `MembershipHelper` exposed as `Members`
- `ProfilingLogger` / Logger

> re: `UmbracoHelper` - we shouldn't need to expose this, as it is typically used to retrieve from the content/media cache, whereas we can use `UmbracoContext.ContentCache.GetById` directly.

---

Any thoughts?

JimBobSquarePants:
Yup. Definitely wise.

jamiepollock:
Late to the party, but this would be awesome. :)

mattbrailsford:
Shall we close this and move the discussion to the PR?

leekelleher:
Onwards to #204 

![Let's go!](http://4.bp.blogspot.com/-IFrDWt43SLk/VcaF1FXt3aI/AAAAAAAAlWg/-KGP08YHKRg/s1600/MontyPythonHolyGrail_018Pyxurz.jpg)

PR #204, opened 2017-01-20
Adding in Base Processor - exposing Umbraco context services

jamiepollock:
As discussed on issue #200.  I've done the work to add UmbracoContext and ApplicationContext in the base `DittoProcessorAttribute`. I used the singleton `UmbracoContext.Current` and `ApplicationContext.Current` as mentioned [here](https://our.umbraco.org/documentation/Reference/Common-Pitfalls/#using-umbracocontext-to-access-applicationcontext) accessing ApplicationContext through UmbracoContext is something you'd want to avoid doing.

I've also reworked the DittoProcessor methods which called `UmbracoContext.Current` to use the new Umbraco property. 

There are a couple of internal statics I didn't touch upon though:

1. https://github.com/leekelleher/umbraco-ditto/blob/8387c47bc5f0a2c01ed5b3299506f91c28bd4234/src/Our.Umbraco.Ditto/ComponentModel/Processors/UmbracoDictionaryAttribute.cs#L61-L76
2. https://github.com/leekelleher/umbraco-ditto/blob/8387c47bc5f0a2c01ed5b3299506f91c28bd4234/src/Our.Umbraco.Ditto/ComponentModel/Processors/UmbracoPickerAttribute.cs#L142-L154

Let me know what you think!


jamiepollock:
I realise there's going to need be a lot more work done in order to get this to work with the tests as UmbracoContext.Current & ApplicationContext.Current are both null. :/

jamiepollock:
Hey folks, struggling with this one. I've been back and forth with how this setup but in the end I took an idea from the core on UmbracoContext access (see notes here: https://github.com/leekelleher/umbraco-ditto/pull/204/commits/bd4c79aaf5ae61a2a83d1f6190e536a545063058).

Happy to rename things, the structure may not be how you guys would see the files being placed and the naming is a tad verbose. This was a bit of a brain racker!

jamiepollock:
I've managed to removed the static membershiphelper but removing the static dictionary helper might be a bit more challenging. I've left it in for now and it can be reviewed later on.

leekelleher:
Hey @jamiepollock, thanks for taking it on and making a start with this! 🎉 

I'm not sure what to make of the `IUmbracoApplicationContextAccessor` yet, (it's the first time I've heard about `IUmbracoContextAccessor` in Umbraco core too). I'll give it some thought.

jamiepollock:
@leekelleher Yea, I know what you mean. The whole reason is to have a testable in for UmbracoContext & ApplicationContext. 

I've been back and forth with this over the weekend after I saw the first push failed I've been trying to get this in. In the end this was the best way I could find. I noticed the Umbraco core was doing something like this in their newer code for Umbraco Identity (https://github.com/umbraco/Umbraco-CMS/search?utf8=%E2%9C%93&q=SingletonUmbracoContextAccessor) and phasing out stuff like plugincontroller. 

I'm not suggesting this is how we should go but I've certainly spent a good amount of time looking over a few options.

I realise adding it to the constructor is out as it'd need to be public to be testable and also then appear in the attribute setup.

It's been a fun one :)

mattbrailsford:
This is one of the biggest drawbacks to the "Context" pattern imo. Testing umbraco v5, which was uber context heavy, was a real pain as you end up spending ages mocking the rabbit hole.

RE the passing in attribute, you could make an internal accessor as I think we make internals visible to the testing library so it should be allowed to be set from the testing lib.

src/Our.Umbraco.Ditto/Ditto.cs
```diff
@@ -135,5 +135,15 @@ public static void RegisterDefaultProcessorType<TProcessorAttributeType>()
...
         }
+
+        /// <summary>
+        /// Registers a global umbraco application context accessor.
+        /// </summary>
+        /// <typeparam name="TUmbracoApplicationContextAccessorType">The type of the accessor.</typeparam>
+        public static void RegisterUmbracoApplicationContextAccessor<TUmbracoApplicationContextAccessorType>()
+            where TUmbracoApplicationContextAccessorType : IUmbracoApplicationContextAccessor, new()
+        {
+            DittoProcessorRegistry.Instance.RegisterDefaultUmbracoApplicationContextAccessorType<TUmbracoApplicationContextAccessorType>();
```
mattbrailsford: Not sure if DittoProcessorRegistery is the right place for this?

src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoProcessorAttribute.cs
```diff
@@ -69,6 +71,56 @@ protected DittoProcessorAttribute()
         internal Type ContextType { get; set; }
 
         /// <summary>
+        /// Returns the MembershipHelper instance
+        /// </summary>
+        protected MembershipHelper Members
+        {
+            get { return new MembershipHelper(Umbraco); }
```
mattbrailsford: I'm wondering if these should be stored on the processor context instead, and then these could maybe be convenience accessors?

src/Our.Umbraco.Ditto/ComponentModel/Processors/DittoProcessorRegistry.cs
```diff
@@ -31,6 +31,11 @@ internal class DittoProcessorRegistry
         private Type defaultProcessorType = typeof(UmbracoPropertyAttribute);
 
         /// <summary>
+        /// The default umbraco application context accessor type for processors, (defaults to `SingletonUmbracoApplicationContextAccessor`).
+        /// </summary>
+        private Type defaultUmbracoApplicationContextAccessorType = typeof(SingletonUmbracoApplicationContextAccessor);
```
mattbrailsford: Thinking this should maybe go directly on Ditto?

jamiepollock:
@mattbrailsford yea, I've not found it _so_ bad. The goal with the accessor was to eliminate the need for any references to UmbracoContext in production code, as if I did put it in then it would cause tests to fail. Hence in the tests we can just setup a mocked up version and be done with it. In theory we could pass in an entirely different UmbracoContext & ApplicationContext constructed setup, somehow. not sure how that'd work mind or if it even should.

On top of this, the parameter nesting of passing the accessor result comes through comes about because we're using extension methods rather than a class we can DI something into.

I might be missing something though. I've tried going down the route of adding in internal properties this weekend and couldn't find the best place to populate those values in the production code which then wouldn't cause tests to fail. Hence coming back around to the accessor.

More than happy to take on edits, I might be close to the trees to see that wood. :)

src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoProcessorAttribute.cs
jamiepollock: Sure, so it'd be Context.Members, etc?

src/Our.Umbraco.Ditto/ComponentModel/Processors/DittoProcessorRegistry.cs
jamiepollock: Yea, I was wondering this too. Happy for it to be moved. :)

src/Our.Umbraco.Ditto/Ditto.cs
jamiepollock: It could be moved to Ditto then, I guess, along with the private field?

src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoProcessorAttribute.cs
mattbrailsford: You could keep these accessors, but maybe the main exposure point is the ProcessorContext, so you could do Context.Members or  just Members (which internally just calls Context.Members). I dunno if this is a good or bad idea, so open to discussion? @leekelleher?

src/Our.Umbraco.Ditto/Ditto.cs
mattbrailsford: Would seem logical, unless we want a DittoUmbracoApplicationContextAccessorRegistry? @leekelleher?

mattbrailsford: Actually, given there can be only 1 type, I'm wondering if this shouldn't just be a static variable on Ditto like ProcessorAttributeTargets and DefaultCacheBy etc. The Register methods are only really relevant if you have multiple types to register.

mattbrailsford: doh! looks like we do have RegisterDefaultProcessorType though, hmmm. Lets keep it how you've got it, but just move it to Ditto including the private property. :)

jamiepollock: Okay. I'll get on that a bit later. :)

Thanks for the feedback, @mattbrailsford!

jamiepollock:
@mattbrailsford I've looked at implementing the code review changes. Only one I've not implemented is helpers into the Context. I wasn't quite sure how to achieve that at this point or if we're going that way.

mattbrailsford:
@jamiepollock no worries. I'm not 100% on that suggestion so happy for it to stay as is and we can always move it later if needs be. It wouldn't change the API on the processors so won't be a big deal to move if we need to.

mattbrailsford:
Sorry, one last thing. I think I'd drop the use of the word "Default" for the UmbracoApplicationAccessor stiff as there can be only one anyway, so it's not like it's a default of a collection (like the Processor one is). Do you think?

jamiepollock:
Sounds like a good idea. It was based, as you can imagine, on the default processor type but we can't specify an accessor on a per model basis. I'll add it in later on when I get back home.

jamiepollock:
@mattbrailsford sorted!

jamiepollock:
I've added an instance of UmbracoHelper to UmbracoDictionaryAttribute.cs. 

I feel this might be something we could add at Processor level as it has useful methods outside of wrapping up Services methods or ContentCache.

jamiepollock:
@mattbrailsford @leekelleher Okay, chaps. I bit the bullet and added in a reference to `UmbracoHelper` along side all the other new additions. Much like Members (`MembershipHelper`) & `UmbracoHelper` referenced elsewhere this is simply called "Umbraco".

This means `UmbracoContext` has been renamed from Umbraco to UmbracoContext.

All references have been updated! I think we're getting closer! :D

src/Our.Umbraco.Ditto/Ditto.cs
```diff
@@ -135,5 +140,26 @@ public static void RegisterDefaultProcessorType<TProcessorAttributeType>()
         {
             TypeDescriptor.AddAttributes(typeof(TObjectType), new TypeConverterAttribute(typeof(TConverterType)));
         }
+
+        /// <summary>
+        /// Registers a global umbraco application context accessor.
+        /// </summary>
+        /// <typeparam name="TUmbracoApplicationContextAccessorType">The type of the accessor.</typeparam>
+        public static void RegisterUmbracoApplicationContextAccessor<TUmbracoApplicationContextAccessorType>()
```
leekelleher: This feels like a mouthful, wondering if this can have a concise name?

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -11,6 +11,7 @@
 
 namespace Our.Umbraco.Ditto
 {
+    using global::Umbraco.Core;
```
leekelleher: We can place the `using` namespaces outside the main namespace. Those left inside are older code. The `global::` part felt unnecessary.

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -191,12 +192,15 @@ public static class PublishedContentExtensions
                 throw new ArgumentException(string.Format("The instance parameter does not implement Type '{0}'", type.Name), "instance");
             }
 
+            // Get the accessor for UmbracoContext & ApplicationContext
+            var umbracoApplicationContextAccessor = (IUmbracoApplicationContextAccessor)Activator.CreateInstance(Ditto.GetUmbracoApplicationContextAccessorType());
```
leekelleher: I think we have an extension method that can create new object instances - I'm lead to believe it has performance optimisations. I'll find out what the method is.

leekelleher: The `umbracoApplicationContextAccessor` variable name feels too verbose... let's consider alternatives 😏

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
```diff
@@ -212,11 +216,11 @@ public static class PublishedContentExtensions
                 if (cacheAttr != null)
                 {
                     var ctx = new DittoCacheContext(cacheAttr, content, type, culture);
-                    return cacheAttr.GetCacheItem(ctx, () => ConvertContent(content, type, culture, instance, processorContexts, onConverting, onConverted));
+                    return cacheAttr.GetCacheItem(ctx, () => ConvertContent(content, type, umbracoApplicationContextAccessor, culture, instance, processorContexts, onConverting, onConverted));
                 }
                 else
                 {
-                    return ConvertContent(content, type, culture, instance, processorContexts, onConverting, onConverted);
+                    return ConvertContent(content, type, umbracoApplicationContextAccessor, culture, instance, processorContexts, onConverting, onConverted);
```
leekelleher: Is there a reason why the `umbracoApplicationContextAccessor` is being passed in as a method parameter from this high up the chain?  I'm assuming it's to do with the lazy-loaded properties, but let's review 👍

tests/Our.Umbraco.Ditto.Tests/DictionaryValueTests.cs
```diff
@@ -18,19 +19,37 @@ public class MyModel
...
+                Resolution.Freeze();
+            }
+        }
+        
+        [TestFixtureTearDown]
+        public void Teardown()
+        {
+            if (Resolution.IsFrozen)
+            {
+                Resolution.Reset();
```
leekelleher: On a complete tangent, I had a problem with the frozen resolution on PR #209 ... and seeing this `Resolution.Reset()` might fix my issue... thanking you! 💯

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
jamiepollock: @leekelleher I wanted to find the highest up point in the chain where we used `UmbracoContext.Current` (where the CurrentCulture is being used) then use then pass the same context through to the extension methods to the place where we used to inject it into the processor attributes.

It didn't seem right to new the class in both cases but I see how passing this through might seem a little odd too. :)

src/Our.Umbraco.Ditto/Ditto.cs
jamiepollock: Hey @leekelleher, I'd like to have a more concise name however it's an class that acts an accessor for both `UmbracoContext` and `ApplicationContext`. I'm at a loss as to what else to call it but I'm happy to hear any suggestions :)

jamiepollock:
I think the main stickler here is the naming of the `UmbracoApplicationContextAccessor`, As mentioned in this commit (https://github.com/leekelleher/umbraco-ditto/pull/204/commits/bd4c79aaf5ae61a2a83d1f6190e536a545063058) I took a cue from the core but that's probably not commonly used code as it was mostly put in for Umbraco Identity.

I don't think UmbracoContextAccessor is quite right as it also accesses ApplicationContext. In the current version of Umbraco we're using IUmbracoContextAccessor is internal but in newer versions it's public.

Happy to rename this to whatever you guys think is best :)

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
leekelleher: @jamiepollock I think it's fine... just wanted to flag it (probably me being pedantic late last night 😉)

@mattbrailsford You cool with this?

mattbrailsford: 👍

leekelleher:
I've created a new feature branch called "feature/umbraco-context", I'll merge this in to there... then we can review/amend accordingly (then get it merged into "develop" branch).

@jamiepollock Thanks again for this, it's very much appreciated! #h5yr 💯

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
JimBobSquarePants: `GetInstance(this Type type)` though internally that calls `Activator.CreateInstance()` if no parameters are passed. We should use it for consistency though.

src/Our.Umbraco.Ditto/Ditto.cs
JimBobSquarePants: @leekelleher `UmbracoPickerAttribute` is the last reference to those `Convert___FromInt` methods. Definintely happy to drop them as `UmbracoPickerAttribute` does everything

src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
leekelleher: @JimBobSquarePants Cool, done - in comment fd44882c2ac5ed5c9cdb68ec25829286278a0eda 👍

src/Our.Umbraco.Ditto/Ditto.cs
leekelleher: @JimBobSquarePants I currently set them as deprecated, (142768ef4a523bbab02bbcd55e63dbfa6fb4339b), but think I may remove them instead now.
@mattbrailsford Any objections?

JimBobSquarePants: Chop chop chop...em

leekelleher: ...and they're gone! 9678d432d52fd9b84eb8d7cfa4b1527c1a17a596

jamiepollock:
Hey guys, is there anything I can do to help this along? Just wondering where we're at. I'm sure you're all super busy at the moment. :)

leekelleher:
@jamiepollock It's the first time I've had chance to look at Ditto in ages.  Reviewing stuff now and we'll see where it's at. 🤘

jamiepollock:
@leekelleher thats okay :) just want to make sure it's in the pipeline. I was happy with the code where I left off but time makes way for uncertainty.

Issue #201, opened 2017-01-10
Custom Processor Contexts only persist a single .As()


drooks:
Per this blog https://jamiepollock.github.io/blog/2016/06/15/getting-started-with-ditto-processors

When passing a custom context as processorContexts to the root .As call, the custom processor is getting null values from the context. It appears that the processorContexts isn't being passed to subsequent .As() calls specifically at DittoFactoryAttribute.cs(line:112).

drooks:
Thanks!

I forked a version and rolled the DittoMultiProcessorContext forward as the processorContexts that I assembled from ContextCache. Problem being it isn't a true stack of processorContexts and includes all cached contexts, which could break if the context type is used differently in the process flow. Needless to say, a fix for this would be fantastic so I can throw away my hack of a fork.

Nicholas-Westby:
Just thought I'd mention the fork for quick reference, which is here: https://github.com/rhythmagency/umbraco-ditto

The commit is here: https://github.com/rhythmagency/umbraco-ditto/commit/52a7868f38c464f7b834d1337aa2098799fe4bc1

jamiepollock:
I believe I had a similar situation when using a custom `DittoFactoryAttribute`. Passing a context in at controller level when we call `model.As<TPoco>(contexts)` but the context didn't pull through to the found POCO class after the factory resolved the type name.

I wasn't sure if this was related, but it might be given the fact that it was being used with Nested Content and the `DittoFactoryAttribute` would trigger another `model.As<TPoco>()` off the `UmbracoPropertyAttribute`. :)

Nicholas-Westby:
@mattbrailsford FYI, I just gave this a test with MyGet and it didn't seem to fix the issue (some of my data was missing). I downgraded back to the custom build of Ditto created by @drooks and the data was fine again.

Here's the MyGet feed I was using: https://www.myget.org/gallery/umbraco-ditto

Latest version, which is 0.11.0-alpha-000685.

mattbrailsford:
Would be useful if you could tell us your model setup so we can replicate
it with a unit test.

Nicholas-Westby:
@mattbrailsford I created a repo that demonstrates the bug: https://github.com/Nicholas-Westby/ditto-context-bug

Clone, run the website from Visual Studio 2015, and the homepage should look like this:

![home](https://cloud.githubusercontent.com/assets/5933222/22517766/1b7653b8-e85f-11e6-9f80-b55411faec12.png)

[...]

PR #207, opened 2017-02-01
Alternative implementation of the ditto chain context

mattbrailsford:
Alternative fix for issue #201

Chain contexts are used to persist the processor context cache across a chain of processors. The curent implementation tries to be clever by maintaining a cache behind the scenes, but this seems to be causing problems. This implementation makes passing of the chain context through the chain more explicit. All processors now have a property of ChainContext and if .As is called inside a processor and you want it to take part in the chain context then you now have to pass the current chain context in

    content.As<Type>(chainContext:ChainContext);

This seems to fix the issues the other implementation had, but forces the passing of the context to be explicit and thus possible for someone to forget. That said, it's sounding like this may be the most logical solution.

Discuss!


jamiepollock:
@mattbrailsford glad it fixes the issue. I'm all for it being being more explicit as to what's going on.

Sorry if I'm being thick here, just trying to make things clear in my head. So, when I'm using multiple processors be it either when decorating a property or in a multi-processor I'd use the ChainContext rather than the Context?

Nicholas-Westby:
@mattbrailsford Sounds like a good solution. I'll give it a test once it's available via MyGet.

To echo what @jamiepollock said, would I also need to use the `chainContext` parameter rather than `processorContext`? That is, on the very first call to `As()` (before the nested calls to `As()`)?

Assuming, of course, that I have a particular context I want to provide.

jamiepollock:
@Nicholas-Westby As this is a PR it won't be built on MyGet. You could always clone the repo and build the package locally using the build.cmd. I have a NuGet feed pointing to a local directory for just this kind of thing.

Just make sure you change your package source to your locally configured nuget feed.

mattbrailsford:
@jamiepollock @Nicholas-Westby the ChainContext is really just an internal thing. It keeps a cache of processor contexts. You shouldn't need to interact with it yourself. Your processor will still be passed the relevant processor context via it's Context property, so you should continue to use that. The ChainContext just allows us to pass "Chain related contextual information" down the entire chain.

Nicholas-Westby:
@mattbrailsford I built this pull request and updated my local and it seemed to work perfectly without any code changes. ✅ 👍

leekelleher:
@mattbrailsford Are you happy for this to be merged in? _or are there still questions around this?_

mattbrailsford:
Merge away :)

PR #210, opened 2017-02-10
Remove UltimatePicker processor

leekelleher:
UltimatePicker was an Umbraco v4/6 property-editor and hasn't been ported to Umbraco v7+
Since our lowest dependency on Umbraco is v7.3.2, we can safely remove this processor.

Unless anyone objects?


JimBobSquarePants:
Kill it with 🔥

PR #211, opened 2017-03-06
WIP Performance improvements.

JimBobSquarePants:
I'm opening this PR as a work-in-progress performance do-over for the library that you can track the progress of. 

I've simplified the BenchmarkDotNet project to make it easier to create tests and have made a couple of changes so far to PropertyInfoInvocations to very slightly improve performance. (I wish i could remove the boxing but I think that's impossible without generics)

I'm going to be focusing mostly on where we are using Linq but I'll keep an eye out elsewhere

### DIFF
--- a/src/Our.Umbraco.Ditto/Common/CachedInvocations.cs
+++ b/src/Our.Umbraco.Ditto/Common/CachedInvocations.cs
@@ -33,6 +33,7 @@ namespace Our.Umbraco.Ditto
         /// <returns>
         /// The <see cref="MethodBaseCacheItem"/> for the given method and type.
         /// </returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         protected static MethodBaseCacheItem GetMethodCacheKey(object type, [CallerMemberName] string memberName = null)
         {
             return new MethodBaseCacheItem(memberName, type);
--- a/src/Our.Umbraco.Ditto/Common/FastPropertyAccessor.cs
+++ b/src/Our.Umbraco.Ditto/Common/FastPropertyAccessor.cs
@@ -1,0 +1,116 @@
+﻿using System;
+using System.Reflection;
+using System.Reflection.Emit;
+using System.Runtime.CompilerServices;
+
+namespace Our.Umbraco.Ditto
+{
+    /// <summary>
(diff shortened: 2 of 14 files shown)


JimBobSquarePants:
I decided to go full hardcore in order to improve the performance of getting/setting properties for our mapped types so I've replaced our `PropertyInfoInvocations` class which used Expression Trees with `FastPropertyAccessor` which gets right down to the metal using dynamic methods +IL

As you can see in the results below we are now running ~4.2x faster than using `PropertyInfo` and ~1.26x faster than our previous `PropertyInfoInvocations` version.

Method |        Mean |    StdErr |     StdDev | Scaled | Scaled-StdDev |
------------------------ |------------ |---------- |----------- |------- |-------------- |
'Static C#' |   1.8323 ns | 0.0190 ns |  0.0735 ns |   1.00 |          0.00 |
'Dynamic C#' |  29.7172 ns | 0.0508 ns |  0.1761 ns |  16.24 |          0.64 |
PropertyInfo | 432.6363 ns | 1.8702 ns |  6.9976 ns | 236.47 |          9.87 |
PropertyDescriptor | 785.7766 ns | 3.7641 ns | 13.0391 ns | 429.49 |         17.98 |
PropertyInfoInvocations | 129.2989 ns | 0.3714 ns |  1.3390 ns |  70.67 |          2.83 |
FastPropertyAccessor | 102.7333 ns | 0.4064 ns |  1.5205 ns |  56.15 |          2.32 |

JimBobSquarePants:
I think this is about as fast as I can get it without making the code unreadable. I think it's good to merge

Nicholas-Westby:
My guess is is that `AggressiveInlining` removes the overhead of a function call. So instead of this:

```cs
var x = sum(1, 2);
```

You get this:

```cs
var x = 1 + 2;
```

Makes the compiled code slightly larger, but for very small functions that are called a bunch, it's a worthwhile tradeoff to improve speed.

JimBobSquarePants:
@Nicholas-Westby Yeah that's right... @leekelleher It's really just a suggestion to the compiler to inline the code, it can still ignore you though. It makes the binary a little larger but on a hotpath it can make quite a difference to performance.

PR #213, opened 2017-03-18
IUmbracoApplicationContextAccessor rename

leekelleher:
With references to PR #212, one of the queries (from PR #204) was the naming of `IUmbracoApplicationContextAccessor`.  How's about renaming it to `IDittoContextAccessor`?


jamiepollock:
I think it makes sense as it's not a processor context.

PR #215, opened 2017-05-30
With OnConverting, should the Content object be mutable?

leekelleher:
Currently with the `OnConverting` event, if you try to set the `Content` property of the `DittoConversionHandlerContext` container object, it doesn't update the original reference of the `IPublishedContent` object.

**My question is, conceptually, should we be able to replace/amend the `IPublishedContent` within an `OnConverting` event?**

---

_I do have an edge-case requirement for this. I'm happy to explain if needed, but wanted to keep the initial question simple._

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/ConversionHandlers/DittoConversionHandler.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/ConversionHandlers/DittoConversionHandler.cs
@@ -121,6 +121,9 @@ namespace Our.Umbraco.Ditto
             this.Model = ctx.Model as TConvertedType;
 
             this.Run(type);
+
+            if (type == DittoConversionHandlerType.OnConverting && this.Content != ctx.Content)
+                ctx.Content = this.Content;
         }
     }
 }
--- a/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
+++ b/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs
@@ -343,7 +343,7 @@ namespace Our.Umbraco.Ditto
 
             // We have the instance object but haven't yet populated properties
             // so fire the on converting event handlers
-            OnConverting(content, type, instance, onConverting);
+            OnConverting(ref content, type, instance, onConverting);
 
             // Process any non lazy properties next
             foreach (var propertyInfo in properties.Where(x => !x.ShouldAttemptLazyLoad()))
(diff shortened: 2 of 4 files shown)


leekelleher:
My understanding of C# reference and value object types gets a bit sketchy here, so please do say if I'm missing something obvious here.

I think the issue is that when the [`DittoConversionHandlerContext` is constructed](https://github.com/leekelleher/umbraco-ditto/blob/0.10.0/src/Our.Umbraco.Ditto/Extensions/PublishedContentExtensions.cs#L610), the reference of the `IPublishedContent` is passed in, but if I attempt to update the `Content` property, the original reference object isn't updated.

I've made this as a PR with relevant amends, mostly so we can discuss via code.

JimBobSquarePants:
>My question is, conceptually, should we be able to replace/amend the IPublishedContent within an OnConverting event?

Absolutely not. The `IPublishedContent` instance should be immutable.

leekelleher:
@JimBobSquarePants LOL, I knew you'd say that! 😉 // @mattbrailsford Any thoughts?

I originally was of the same opinion, then was trying to work a tricky problem/scenario - which brought me to the `OnConverting` event... and that `DittoConversionHandlerContext`'s `Content` property having a public setter got me thinking, and to asking this question.  _Which would have solved my particular problem, but yes, would be code smell._

---

On the topic of the `OnConverting` event... I'm now struggling to see the relevance of it.  I'm not sure what it could be used for.  I guess we added it to have symmetry with the `OnConverted` event?

I'll close off this PR before the end of the week, (giving @mattbrailsford a chance to comment on it).  At the very least, I'd like to tight up the `DittoConversionHandlerContext` properties, making the setters internal.

mattbrailsford:
Sorry Lee, but I agree with James on this one. I don't think you should be
able to change the context content and so would agree with you that we need
to make the setters internal.

leekelleher:
@mattbrailsford Totally fine, was more to bounce the idea.

---

Just to justify myself, mostly so you both don't think I've gone bonkers.  I'm prototyping an idea of an IPublishedContent "superset" - where an content node could inherit property-values from another content node, (_@mattbrailsford - for reference, it's the master products on CBG_).

[See my gist of `SupersetPublishedContent`](https://gist.github.com/leekelleher/91022cd1f5ef7b20a0d3623161f2d6f2)... then it's how & where to construct this superset object, especially since I'm using DittoView.  So the `OnConverting` event kinda made sense.

> On a side note, I did look at using Umbraco core's ModelFactory, but there are tight restrictions on that - as models must inherit from `IPublishedContentExtended`, which itself is still marked as internal, urgh. 🤢  ... _I digress._

JimBobSquarePants:
😈 Happy to be predictable in that manner.

Truth be told all our models should be immutable, (though that's impossible of course), as far as I'm aware that's the plan with nuCache + models builder come v8.

PR #218, opened 2017-05-31
DittoConversionHandler - replace Content/Model properties with a Context property

leekelleher:
**Forewarning: Breaking Change**

@mattbrailsford You're gonna hate on me for this! :laughing:

After adding the `Culture` property to `DittoConversionHandlerContext ` (in PR #216). I'd later noticed that I'd missed adding it on `DittoConversionHandler`.

Which got me thinking, why are we duplicating properties here? Wouldn't it be better to access the Content, Model, ModelType (and now Culture) via a Context property?

> Which is conceptually the same as we do in Processors - and the method-level `DittoOnConvertedAttribute` - those properties hang off a context object.

Of course, this being a breaking change, I don't take it lightly, hence seeking peer approval. ❤️


JimBobSquarePants:
This is a good thing. It's always wise to pass a since context object though event handlers. 👍

PR #223, opened 2017-08-31
Post processors - Adds methods to register/deregister

leekelleher:
@mattbrailsford What are you're thoughts on the implementation?
A basic list with the register/unregister methods, wrapping the list's add/insert/remove methods.

I'd looked at making the entire post-processors list replaceable, but `RecursiveDittoAttribute` and `TryConvertToAttribute` are marked as internal, so that wasn't ideal, (and I don't want to change those to public).

For the unregister one, it gives me the option to exclude the `HtmlString` processor when I don't need it.

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/Processors/DittoProcessorRegistry.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/Processors/DittoProcessorRegistry.cs
@@ -11,9 +11,9 @@ namespace Our.Umbraco.Ditto
     internal class DittoProcessorRegistry
     {
         /// <summary>
-        /// The cache for storing handler information.
+        /// The cache for storing processors associated by type.
         /// </summary>
-        private static readonly Dictionary<Type, List<DittoProcessorAttribute>> Cache = new Dictionary<Type, List<DittoProcessorAttribute>>();
+        private static readonly Dictionary<Type, List<DittoProcessorAttribute>> ProcessorCache = new Dictionary<Type, List<DittoProcessorAttribute>>();
 
         /// <summary>
         /// Static holder for singleton instance.
@@ -21,9 +21,9 @@ namespace Our.Umbraco.Ditto
         private static readonly Lazy<DittoProcessorRegistry> InternalInstance = new Lazy<DittoProcessorRegistry>(() => new DittoProcessorRegistry());
 
         /// <summary>
-        /// The lock object to make Cache access thread safe
+        /// The lock object to make ProcessorCache access thread safe.
         /// </summary>
-        private static readonly object CacheLock = new object();
+        private static readonly object ProcessorCacheLock = new object();
 
         /// <summary>
         /// The default processor type, (defaults to `UmbracoProperty`).
@@ -93,14 +93,14 @@ namespace Our.Umbraco.Ditto
         {
             var objType = typeof(TObjectType);
 
-            lock (CacheLock)
+            lock (ProcessorCacheLock)
             {
-                if (!Cache.ContainsKey(objType))
+                if (!ProcessorCache.ContainsKey(objType))
                 {
-                    Cache.Add(objType, new List<DittoProcessorAttribute>());
+                    ProcessorCache.Add(objType, new List<DittoProcessorAttribute>());
                 }
 
-                Cache[objType].Add(instance);
+                ProcessorCache[objType].Add(instance);
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Our.Umbraco.Ditto/ComponentModel/Processors/DittoProcessorRegistry.cs
```diff
@@ -31,6 +31,17 @@ internal class DittoProcessorRegistry
         private Type DefaultProcessorType = typeof(UmbracoPropertyAttribute);
 
         /// <summary>
+        /// The default post-processors.
+        /// </summary>
+        private List<DittoProcessorAttribute> DefaultPostProcessorAttributes = new List<DittoProcessorAttribute>()
```
mattbrailsford: I know nothing about thread safety, but given we protect the Cache collection, I wonder if we need to do the something with the list. I'm not sure if there is anything baked in we can used, or whether we'd need to use another lock, but worth thinking about.

src/Our.Umbraco.Ditto/ComponentModel/Processors/DittoProcessorRegistry.cs
```diff
@@ -141,6 +148,36 @@ public IEnumerable<DittoProcessorAttribute> GetRegisteredProcessorAttributesFor(
...
+            {
+                this.DefaultPostProcessorAttributes.Insert(position, processor);
+            }
+        }
+
+        /// <summary>
+        /// Unregisters a processor attribute from the default set of post-processor attributes.
+        /// </summary>
+        /// <typeparam name="TProcessorAttributeType"></typeparam>
+        public void UnregisterPostProcessorAttribute<TProcessorAttributeType>()
```
mattbrailsford: DeregisterProcessorAttribute?

src/Our.Umbraco.Ditto/Ditto.cs
```diff
@@ -161,5 +161,26 @@ public static Type GetContextAccessorType()
...
+            where TProcessorAttributeType : DittoProcessorAttribute, new()
+        {
+            DittoProcessorRegistry.Instance.RegisterPostProcessorAttribute<TProcessorAttributeType>(position);
+        }
+
+        /// <summary>
+        /// Unregisters a processor attribute from the default set of post-processor attributes.
+        /// </summary>
+        /// <typeparam name="TProcessorAttributeType"></typeparam>
+        public static void UnregisterPostProcessorAttribute<TProcessorAttributeType>()
```
mattbrailsford: DeregisterProcessorAttribute?

src/Our.Umbraco.Ditto/ComponentModel/Processors/DittoProcessorRegistry.cs
leekelleher: OK cool, I'll add a lock in.  I'm wondering if we can reuse `CacheLock` for it?

...and for when adding/inserting, do you think we should check if the processor type has already been added? (prevent duplicates)

src/Our.Umbraco.Ditto/ComponentModel/Processors/DittoProcessorRegistry.cs
leekelleher: Agreed, with the thinking being: "deregister" is the act and "unregister" is the state

src/Our.Umbraco.Ditto/ComponentModel/Processors/DittoProcessorRegistry.cs
mattbrailsford: I don't think  you should use the same lock as you could end up holding up the other function, so probably go with another lock.

I'm wondering if there is ever the need to have the same processor more than once? Could be possible, but can't think of a use case atm. Happy either way.

leekelleher: @mattbrailsford Do you think this should a `readonly` field? and what about making it `static`? given that the processor cache is static.

mattbrailsford: Good point, yea, it should be and in fact, I think DefaultProcessorType should be static too (I know it's not a change related to this PR though so can be don't direct in develop branch, just mentioning it)

leekelleher: I'll do the `DefaultProcessorType` one as a separate PR.  Can't be `readonly` though, as it can be replaced.

leekelleher:
@mattbrailsford PR updated, as per review comments.

PR #224, opened 2017-09-04
DefaultProcessorType - makes static and adds lock

leekelleher:
As discussed on PR #223 review: https://github.com/leekelleher/umbraco-ditto/pull/223#discussion_r136792514


src/Our.Umbraco.Ditto/ComponentModel/Processors/DittoProcessorRegistry.cs
```diff
@@ -108,7 +116,10 @@ public Type GetDefaultProcessorType(Type objectType)
                 return attr.ProcessorType;
             }
 
-            return this.DefaultProcessorType;
+            lock (DefaultProcessorTypeLock)
```
mattbrailsford: You only need a lock for writing, not for reading.

PR #225, opened 2017-09-04
Fix proxy generation

JimBobSquarePants:
So it turns out setting properties for proxied instances was broken for value types.

We got away with it because we actually had two instances of the mapped type (one proxied and one not) on the go at one time. (obviously not efficient) and were setting the property on the hidden copy.

This PR fixes that and also changes the interceptor to only attempt to intercept the lazy properties (as removing the copy would create an infinite recursion of interception)

We also don't now attempt to create a proxy from a non-null instance as that would simply replace the instance.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,4 @@
-image: Visual Studio 2017
-
+os: Visual Studio 2015
 
 # version format
 version: 0.11.1.{build}
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,5 @@
-os: Visual Studio 2015
+image: Visual Studio 2017
+
 
 # version format
 version: 0.11.1.{build}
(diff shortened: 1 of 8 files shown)


JimBobSquarePants:
`Reflection.Emit` is hard, very hard but I understand it enough to write something simple-ish like this. Anything more complicated though would require a lot of swearing!

Protip: If you ever end up giving this kind of code a go Teleriks Just Decompile is leagues ahead of Jetbrains DotPeek. Gives you tons to work with viewing output and the Peverify.exe tool in the Windows SDK tells you exactly what is wrong when you mess up.

I added tests also with a demo very similar to what the reflection emit code actually produces so we know it works also.

leekelleher:
Merged in, thanks @JimBobSquarePants! 🎉

PR #232, opened 2017-10-15
Attributed Type Resolver

leekelleher:
Adds a resolver class to provide a lookup for specific Ditto attributes for a given type.
The thinking behind this is that we have a few places where we check for a custom attribute, e.g. the `UmbracoProperties` attribute inside the `UmbracoProperty` processor.
Given that this is called for each object's property, it has the potential of being queried a lot of times.

Of course this adds time to app startup, but then it's a once-off operation.

Would like to know your thoughts on this approach.

### DIFF
--- a/src/Our.Umbraco.Ditto/Bootstrapper.cs
+++ b/src/Our.Umbraco.Ditto/Bootstrapper.cs
@@ -1,0 +1,29 @@
+﻿using Umbraco.Core;
+
+namespace Our.Umbraco.Ditto
+{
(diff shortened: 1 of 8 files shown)


leekelleher:
@mattbrailsford Following up on our conversation about this, I tried to look at an alternative way, so we could avoid all the type lookups at app-startup, but my prototype didn't work out. I ran into issues with null values in the ConcurrentDictionary - otherwise if we'd end up checking if the attribute exists every time.

I'll merge in this PR and we can refactor as needed, (it's all currently marked as internal).

PR #233, opened 2018-01-09
Deprecate PropertyDescriptor in Contexts

leekelleher:
The reason we're using the `PropertyDescriptor` is due to Ditto original use of TypeConverters.
When we rolled up the TypeConverters and ValueResolvers, we'd kept `PropertyDescriptor` available.

However the use of reflection in Ditto's internals, there is an additional call made to exclusively get the `PropertyDescriptor`. If we switch this to use the `PropertyInfo`, (which we already have a local reference to), then this can be used to access any metadata about the property itself.

So not to make this a rough breaking-change, I'd left the `PropertyDescriptor` property available on the Contexts, so not to completely ruin the developer experience.  I added an obsolete/warning message saying that it would be removed in a future release.

Does anyone foresee any downsides with this approach?

### DIFF
--- a/src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoProcessorAttribute.cs
+++ b/src/Our.Umbraco.Ditto/ComponentModel/Attributes/DittoProcessorAttribute.cs
@@ -196,7 +196,7 @@ namespace Our.Umbraco.Ditto
             this.Context = context;
             this.ChainContext = chainContext;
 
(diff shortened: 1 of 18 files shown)


mattbrailsford:
Sounds good to me 👍

JimBobSquarePants:
PropertyDescripters are slow, ditch em when you can.

leekelleher:
Thanks guys, merged it in!

PR #243, opened 2018-12-16
Feature/undefer

Nicholas-Westby:
This is to potentially address some gremlins I've noticed with Ditto in which it sometimes runs slow and sometimes seems to run the same code multiple times.

I haven't tested this yet. For now, I'm curious if you think this will have any impact. If I get a free moment, I'll test this on some websites I've been working with to see if if affects anything.

This relates to this discussion: https://our.umbraco.com/packages/developer-tools/ditto/ditto-feedback/94330-deferred-execution-causing-repeat-execution

I also noticed some typos that I fixed in this PR.

### DIFF
--- a/src/Our.Umbraco.Ditto/Common/FastPropertyAccessor.cs
+++ b/src/Our.Umbraco.Ditto/Common/FastPropertyAccessor.cs
@@ -6,8 +6,8 @@ using System.Runtime.CompilerServices;
 namespace Our.Umbraco.Ditto
 {
     /// <summary>
-    /// Provides a way to set a properties value using a combination of dynamic methods and IL generation.
-    /// This is much faster 4.3x than <see cref="M:PropertyInfo.SetValue"/> as it avoids the normal overheads of reflection.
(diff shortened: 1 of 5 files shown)


Nicholas-Westby:
BTW, some extra context for the changes made:

* I added ToArray to the .As extension method that operations on IEnumerables. This is called from RecursiveDittoAttribute, which I think is called on every property by default. This may affect a large portion of properties that are collections.
* I added ToArray to DittoFactoryAttribute, which is used in scenarios where you may be creating lists of items (I'm pretty sure I use this indirectly anytime I map a collection of widgets, which is the primary use case I have for Ditto).

leekelleher:
Thanks @Nicholas-Westby! I'll try to take a look over it this week.

leekelleher (approved):
Firstly, thanks for correcting the typos, very much appreciated!

I've ran the performance benchmark tests with & without the `.ToArray()` calls, the different is negligible. I'm happy to merge this PR in. Thanks!

Nicholas-Westby:
Awesome!

